PR #4, opened 2017-05-25
Add armhf dockerfiles for deb building

seemethere:
Signed-off-by: Eli Uriegas <seemethere101@gmail.com>


andrewhsu:
As you mentioned, the `CHOWN` variable in `deb/Makefile` needs to take into account running on an armhf machine. Logic to switch off of the `uname -m` output would help.

seemethere:
Added with f3cffac

andrewhsu:
LGTM

PR #7, opened 2017-05-30
deb: add sudo users to the docker group

tych0:
When users install docker-ce, the first thing they have to do is add
themselves to the docker group (see below). Let's instead do this for them, so
a simple `newgrp docker` is all that's required. (See the patch notes for a sample terminal session for why this is helpful)

Note: since having access to the docker API is effectively equivalent to root
access, we should only do this for users in the sudo group.

Also, this probably needs a update to the docs and such. Where should I send those PRs?

### DIFF
--- a/deb/common/docker-ce.postinst
+++ b/deb/common/docker-ce.postinst
@@ -7,6 +7,11 @@ case "$1" in
 			if ! getent group docker > /dev/null; then
 				groupadd --system docker
 			fi
+
+			# Add each sudo user to the docker group
+			for u in $(getent group sudo | sed -e "s/^.*://" -e "s/,/ /g"); do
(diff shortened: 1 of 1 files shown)


andrewhsu:
Although this PR is convenient if a user of the docker-ce package would want to run docker commands uninhibited, I feel possibly the safest option is to let the administrator choose to add all users from the unix `sudo` group to the unix `docker` group.

Besides, not all unix user accounts that are desirable for administration may be added to the unix group; they may be added to the `/etc/sudoers` file only. Would like to have `docker-ce` package not strongly-tied to the operations of the `sudo` package.

seemethere:
Going to close this PR, will reopen if we feel that this is something we would like to add.

PR #9, opened 2017-05-31
Update deb dockerfiles to $(uname -m) names

seemethere:
The dockerfiles were reflecting what was found in the `go env HOSTARCH` variable which doesn't work everywhere.

Now they reflect `uname -m` which should be more portable.


andrewhsu:
LGTM

PR #10, opened 2017-06-01
Add reflective alpine images to other components

seemethere:
Was running into issues where the alpine images were failing on armhf / s390x machines


rpm/Makefile
```diff
@@ -1,11 +1,12 @@
 ARCH=$(shell uname -m)
+ALPINE:=$(shell $(CURDIR)/../detect_alpine_image)
```
andrewhsu: would prefer this var be called `ALPINE_IMG`

andrewhsu: is `detect_alpine_image` part of this PR?

seemethere: No It was included as part of: https://github.com/docker/docker-ce-packaging/pull/4

seemethere: Change to ALPINE_IMG as of c423146

andrewhsu:
LGTM

PR #20, opened 2017-06-19
restart docker service if needed for rpm upgrade

andrewhsu:
Addresses issue moby/moby#33688: Container can not start after docker engine upgrade from 17.03.1-ce to 17.06.0-ce-rc3

This PR will stop the docker service if it is running in an upgrade scenario (docker service is never started on a fresh install as per RPM behaviour conventions). It will then restart the service after the new files are installed so that the new daemon is used to launch the service.

For reference, RPM spec file scriptlet ordering https://fedoraproject.org/wiki/Packaging:Scriptlets

To build a centos rpm:
```
$ make -C rpm ENGINE_DIR=/path/to/engine/code CLI_DIR=/path/to/cli/code centos-7 # rpm will be in rpm/rpmbuild
```

Signed-off-by: Andrew Hsu <andrewhsu@docker.com>

### DIFF
--- a/rpm/centos-7/docker-ce.spec
+++ b/rpm/centos-7/docker-ce.spec
@@ -154,6 +154,20 @@ install -p -m 644 engine/contrib/syntax/nano/Dockerfile.nanorc $RPM_BUILD_ROOT/u
 /usr/share/vim/vimfiles/syntax/dockerfile.vim
 /usr/share/nano/Dockerfile.nanorc
 
+%pre
+if [ $1 -gt 0 ] ; then
+    # package upgrade scenario, before new files are installed
+
+    # clear any old state
+    rm -f %{_localstatedir}/lib/rpm-state/docker-is-active > /dev/null 2>&1 || :
+
+    # check if docker service is running
+    if systemctl is-active docker > /dev/null 2>&1; then
+        systemctl stop docker > /dev/null 2>&1 || :
+        touch %{_localstatedir}/lib/rpm-state/docker-is-active > /dev/null 2>&1 || :
+    fi
+fi
+
 %post
 %systemd_post docker
 if ! getent group docker > /dev/null; then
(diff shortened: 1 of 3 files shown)


rpm/centos-7/docker-ce.spec
```diff
@@ -154,7 +154,30 @@ install -p -m 644 engine/contrib/syntax/nano/Dockerfile.nanorc $RPM_BUILD_ROOT/u
...
+%pre
+if [ $1 -gt 1 ] ; then
+    # package upgrade scenario, before new files are installed
+
+    # clear any old state
+    rm -f %{_localstatedir}/lib/rpm-state/docker-is-active > /dev/null 2>&1 || :
+
+    # check if docker service is running
+    if systemctl is-active docker > /dev/null 2>&1; then
+        systemctl stop docker > /dev/null 2>&1 || :
```
tiborvass: you can redirect both stdout and stderr with `&> /dev/null`

tiborvass:
Out of curiosity, why silent the output of the scripts?

andrewhsu:
@tiborvass just following RPM conventions, e.g. from the `systemd.macros` scriptlets for RPM spec files:

https://github.com/systemd/systemd/blob/a4dde27d73c1d8219c0cc6390ef8b6e15b347edb/src/core/macros.systemd.in#L52

andrewhsu:
@tianon curious to know if this PR looks sane to you?

tianon:
Seems mostly reasonable -- the reason we didn't do this before is that stopping the daemon will stop all running containers (unless the user has specifically done extra work to prevent that).

rpm/centos-7/docker-ce.spec
tiborvass: Nevermind, this is a bashism.

andrewhsu:
@tianon looks like existing behaviour of docker-ce rpm package upgrade is to restart the systemd docker service if it was already running. i've verified such behaviour when upgrading from docker-ce 17.03.1 to docker-ce 17.05.0

tianon:
Huh, ok!  Sounds good! :+1:

tiborvass:
@tianon for context: what's weird is that upgrading from old rpm to new rpm is where the restart issue happens and I'm not clear why.

tiborvass:
LGTM

tiborvass:
True actual LGTM

PR #21, opened 2017-06-19
Prefixes documentation files per component

seemethere:
Since we pull from multiple components to build docker-ce now it makes sense to have all of the related files associated with each component packaged into RPMS.

This will prefix all files that were previously included in the RPM with the component from which they originate.

End result will look like so:

screen shot 2017-06-19 at 1 19 47 pm

### DIFF
--- a/rpm/centos-7/docker-ce.spec
+++ b/rpm/centos-7/docker-ce.spec
@@ -130,11 +130,11 @@ install -d $RPM_BUILD_ROOT/usr/share/nano
 install -p -m 644 engine/contrib/syntax/nano/Dockerfile.nanorc $RPM_BUILD_ROOT/usr/share/nano/Dockerfile.nanorc
 
 mkdir -p build-docs
-for engine_file in engine/AUTHORS engine/CHANGELOG.md engine/CONTRIBUTING.md engine/LICENSE engine/MAINTAINERS engine/NOTICE engine/README.md; do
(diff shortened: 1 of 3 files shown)


rpm/centos-7/docker-ce.spec
```diff
@@ -129,9 +129,18 @@ install -p -m 644 engine/contrib/syntax/vim/syntax/dockerfile.vim $RPM_BUILD_ROO
 install -d $RPM_BUILD_ROOT/usr/share/nano
 install -p -m 644 engine/contrib/syntax/nano/Dockerfile.nanorc $RPM_BUILD_ROOT/usr/share/nano/Dockerfile.nanorc
 
+mkdir -p build-docs
+for engine_file in engine/AUTHORS engine/CHANGELOG.md engine/CONTRIBUTING.md engine/LICENSE engine/MAINTAINERS engine/NOTICE engine/README.md; do
+    cp "$engine_file" "build-docs/engine-$(basename $engine_file)"
+done
+for cli_file in cli/LICENSE cli/MAINTAINERS cli/NOTICE cli/README.md; do
```
andrewhsu: No need to depend on `basename`:

```
for cli_file in LICENSE MAINTAINERS NOTICE README.md; do
  cp "cli/$cli_file" "build-docs/cli-$cli_file"
done
```

seemethere: Fixed per 8251d8f

PR #22, opened 2017-06-19
deb pkg stop service before upgrade, start after upgrade

andrewhsu:
Similar to PR #20 **restart docker service if needed for rpm upgrade**, this PR addresses issue moby/moby#33688 **Container can not start after docker engine upgrade from 17.03.1-ce to 17.06.0-ce-rc3** for deb packages.

This PR will stop the docker service if it is running in an upgrade scenario. It will then start the service after the new files are installed so that the new daemon is used to launch the service. Unlike rpm packages, installing the docker-ce deb package is always expected to finish by starting the service.

For reference, deb pkg script event hooks: https://www.debian.org/doc/manuals/debian-faq/ch-pkg_basics.en.html#s-maintscripts

To build a xenial deb pkg:
```
$ make -C rpm ENGINE_DIR=/path/to/engine/code CLI_DIR=/path/to/cli/code ubuntu-xenial # deb will be in deb/debbuild
```

Signed-off-by: Andrew Hsu <andrewhsu@docker.com>

### DIFF
--- a/deb/common/docker-ce.postinst
+++ b/deb/common/docker-ce.postinst
@@ -9,6 +9,9 @@ case "$1" in
 			fi
 		fi
 		;;
+	upgrade)
+		invoke-rc.d docker start || exit $?
+		;;
 	abort-*)
 		# How'd we get here??
 		exit 1
--- a/deb/common/docker-ce.preinst
+++ b/deb/common/docker-ce.preinst
@@ -1,0 +1,16 @@
+#!/bin/sh
+set -e
+
+case "$1" in
+	upgrade)
+		invoke-rc.d docker stop || exit $?
+		;;
+	abort-*)
+		# How'd we get here??
+		exit 1
+		;;
+	*)
+		;;
+esac
+
+#DEBHELPER#


tych0:
+1, thanks!

deb/common/docker-ce.preinst
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+set -e
+
+case "$1" in
+	upgrade)
+		invoke-rc.d docker stop || exit $?
```
tiborvass: @andrewhsu what if docker was already stopped?

deb/common/docker-ce.postinst
```diff
@@ -9,6 +9,9 @@ case "$1" in
 			fi
 		fi
 		;;
+	upgrade)
+		invoke-rc.d docker start || exit $?
```
tiborvass: @andrewhsu what is the expected behavior when docker was not running?

tianon:
https://manpages.debian.org/stretch/debhelper/dh_installinit.1.en.html is probably a useful link to get an idea of the code we're getting automatically via `#DEBHELPER#`, and what knobs are available for tweaking it out-of-the-box (would go in `rules` under `common` on our `dh_installinit` invocation; although from what I can tell, this should already be the current behavior, so I'm not sure what's going wrong :disappointed:).

tiborvass:
@tianon if you get docker from http://apt.dockerproject.org/repo/pool/main/d/docker-engine/ and then update to https://download.docker.com/linux/ such as https://download.docker.com/linux/ubuntu/dists/xenial/pool/test/amd64/docker-ce_17.06.0~ce~rc4-0~ubuntu_amd64.deb then it fails.

BUT, if you got docker from, say https://download.docker.com/linux/ubuntu/dists/xenial/pool/stable/amd64/docker-ce_17.03.1~ce-0~ubuntu-xenial_amd64.deb, it's the same `docker version`, but it was packaged by a different script (assuming the one in this repo). The old packaging was from docker/docker's contrib/

deb/common/docker-ce.preinst
andrewhsu: the result of that command will always be exit code 0

deb/common/docker-ce.postinst
andrewhsu: expected behaviour is to have docker service started after completion of install of a docker-ce deb package. this is what happens on a new install of a docker-ce deb package.

andrewhsu:
Closing this in favour of PR #24 **also have deb pkg conflict docker-engine** because tested with that one and it solves the problem.

PR #28, opened 2017-06-22
Fix paths for completion scripts

thaJeztah:
The completion scripts have been moved to the docker/cli repository, so should be taken from that directory.

Note that the docker/docker-ce repository is currently in a mixed state; these two pull requests should be included there;

https://github.com/docker/cli/pull/189
https://github.com/moby/moby/pull/33665

ping @tiborvass @andrewhsu PTAL


andrewhsu:
LGTM

PR #30, opened 2017-07-07
Add LICENSE (copy of docker/docker's Apache-2.0 file)

tianon:
This repository was missing a license, but since it's been created mostly from bits of `github.com/docker/docker`, I figured that license was probably appropriate for this top-level (since the individual vendored directories should have their `LICENSE` files intact anyhow).

A quick `diff` to show that it's 100% identical:

```console
$ diff -u \
  <(curl -fsSL 'https://github.com/docker/docker/raw/master/LICENSE') \
  <(curl -fsSL 'https://github.com/tianon/docker-ce-packaging/raw/license/LICENSE')
$ 
```

See also https://github.com/docker/docker-ce/pull/103. :+1:

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,191 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        https://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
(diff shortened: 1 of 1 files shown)


friism:
Thanks for sending this PR. I don't think we should add a permissive license to this repo, however. The code in this repo is meant for the sole use of Docker Inc. to build packages for the Docker Community Edition products.

If this code had a permissive license, that might be misunderstood to mean that it's ok for others to build and distribute "Docker" packages (it's not ok, "Docker" is protected by a trademark owned by Docker Inc.).

Anyone interested in building something out of the open source components that go into Docker should use the Moby project, not this code: https://github.com/moby/moby

This packaging code is made available to provide transparency for Docker CE users, and to help them troubleshoot and report issues with Docker CE packages.

tianon:
To put that another way, this isn't a "license" to use the Docker trademark, but merely governs how this _code_ is allowed to be re-used, and in this case, is only reflecting the reality of the license this code is already under (due to it having been previously released with this license).

friism:
@tianon I recognize that the upstream licenses already apply to this code, and I recognize that not having the correct license here is unhygienic. You're also correct that the trademark status of the artifacts produced by this code is strictly speaking an orthogonal concern.

Still, I think it's preferable to not include a license in this repo.

AkihiroSuda:
> Still, I think it's preferable to not include a license in this repo.

https://github.com/docker/docker-ce-packaging/blob/6c1f016b124c8f57444cce118266a5b5060ac6aa/README.md

README says "This repo contains the **open source** scripts for packaging Docker CE products." 🤔

thaJeztah:
let's merge

PR #32, opened 2017-07-13
[ppc64le] add deb support for ubuntu-xenial

tophj-ibm:
Adds ubuntu-xenial as a make deb target for ppc64le.

I've tested this with 17.06, but can't get a successful build with
the upstream cli repo because of https://github.com/docker/for-linux/issues/54.

Signed-off-by: Christopher Jones <tophj@linux.vnet.ibm.com>

### DIFF
--- a/deb/ubuntu-xenial/Dockerfile.ppc64le
+++ b/deb/ubuntu-xenial/Dockerfile.ppc64le
@@ -1,0 +1,28 @@
+FROM ppc64le/ubuntu:xenial
+
+RUN apt-get update && apt-get install -y apparmor bash-completion btrfs-tools build-essential cmake curl ca-certificates debhelper dh-apparmor dh-systemd git libapparmor-dev libdevmapper-dev libltdl-dev libseccomp-dev pkg-config vim-common libsystemd-dev --no-install-recommends && rm -rf /var/lib/apt/lists/*
+
+ENV GO_VERSION 1.8.3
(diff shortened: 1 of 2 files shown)


rpm/Makefile
```diff
@@ -1,11 +1,12 @@
+ALPINE:=$(shell $(CURDIR)/../detect_alpine_image)
```
tophj-ibm: this isn't needed by this PR (and I'll make a new one if you want) but fixes non-x86 rpm builds

seemethere: For reference this relates to https://github.com/docker/docker-ce-packaging/pull/10

seemethere: CC @andrewhsu

tophj-ibm: rebased and remove, thanks!

tophj-ibm:
np @andrewhsu. I recently did our own internal release for 17.06.1-ce, and 14.04, 16.10, and 17.04 should all be fine too.

tophj-ibm:
FYI in case you guys haven't hit this already, there is a bug in canonical's cloud image for 14.04 where it defaults to the wrong apt-sources for security updates on non-x86_64 https://bugs.launchpad.net/cloud-images/+bug/1711735.

s390x doesn't have a trusty, but armhf, arm64, and ppc64le should be affected by this with the latest trusty docker images

PR #33, opened 2017-07-13
[s390x] add seccomp buildtag for ubuntu-yakkety

tophj-ibm:
Adds the seccomp buildtag for ubuntu-yakkety on s390x

Note: s390x requires libseccomp-dev 2.3.1 at the moment which is only available
only yakkety and later. This change is also consistent with the s390x yakkety
deb dockerfile in moby/moby.

Signed-off-by: Christopher Jones <tophj@linux.vnet.ibm.com>


andrewhsu:
LGTM

PR #36, opened 2017-07-21
Have docker cli be compiled in GOPATH

seemethere:
manpage generation was failing for docker cli when being run under
aarch64: 

https://ci.qa.aws.dckr.io/job/docker/job/release-packaging/view/change-requests/job/PR-41/5/execution/node/360/log/

The way to remedy this is to compile the cli and the manpages while in
the GOPATH so that dependencies found in the vendor folder are
discovered by the go build tools.

### DIFF
--- a/deb/common/rules
+++ b/deb/common/rules
@@ -12,7 +12,7 @@ override_dh_gencontrol:
 
 override_dh_auto_build:
 	cd engine && ./hack/make.sh dynbinary
-	LDFLAGS='' make -C cli VERSION=$(VERSION) GITCOMMIT=$(DOCKER_GITCOMMIT) dynbinary manpages
+	LDFLAGS='' make -C /go/src/github.com/docker/cli VERSION=$(VERSION) GITCOMMIT=$(DOCKER_GITCOMMIT) dynbinary manpages
 
 override_dh_auto_test:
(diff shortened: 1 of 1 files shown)


seemethere:
CC @andrewhsu

deb/common/rules
```diff
@@ -12,7 +12,7 @@ override_dh_gencontrol:
 
 override_dh_auto_build:
 	cd engine && ./hack/make.sh dynbinary
-	LDFLAGS='' make -C cli VERSION=$(VERSION) GITCOMMIT=$(DOCKER_GITCOMMIT) dynbinary manpages
+	cd /go/src/github.com/docker/cli && LDFLAGS='' make VERSION=$(VERSION) GITCOMMIT=$(DOCKER_GITCOMMIT) dynbinary manpages
```
andrewhsu: instead of `cd`, use the `-C` option of `make` because that's what it does:

> Change to directory dir before reading the makefiles or doing anything else.

seemethere: Tried with `make -C` and it didn't seem to function correctly

andrewhsu:
LGTM

andrewhsu:
Nuts, sorry. Trigger happy. Another PR to follow to do it in a way that works.

PR #39, opened 2017-07-22
Add content verification to the plugin installation

rogaha:
- remove the `tgz` artifacts and use `docker plugin install` instead

Signed-off-by: Roberto Gandolfo Hashioka <roberto_hashioka@hotmail.com>

### DIFF
--- a/common/docker-telemetry-plugin
+++ b/common/docker-telemetry-plugin
@@ -1,0 +1,68 @@
+#!/usr/bin/env sh
+
+BASE=/var/lib/docker
+PLUGIN_NAME=docker/telemetry
+EDITION_INFO=$BASE/.edition-information
+
+# make sure that docker is running
+DOCKER_SOCKET=
+if ! printf "%s" "$DOCKER_OPTS" | grep -qE -e '-H|--host'; then
+        DOCKER_SOCKET=/var/run/docker.sock
+else
+        DOCKER_SOCKET=$(printf "%s" "$DOCKER_OPTS" | grep -oP -e '(-H|--host)\W*unix://\K(\S+)' | sed 1q)
+fi
+
+if [ -n "$DOCKER_SOCKET" ]; then
+        while ! [ -e "$DOCKER_SOCKET" ]; do
+                sleep 0.1
+        done
+else
+        # docker socket not found, exiting...
+        exit
+fi
+
+# check if the file exists
+if ! [ -e "$EDITION_INFO" ]; then
+        # edition metadata not found, exiting...
+        exit
+fi
+
+# Create the engine UUID, if it doesn't exist
+# This will be eventually handled by the docker engine
+# ref. https://github.com/moby/moby/issues/33356
+if [ -f $BASE/engine_uuid ]; then
+    UUID=$(cat $BASE/engine_uuid)
+fi
+if [ -z ${UUID+x} ]; then
+    UUID=$(cat /proc/sys/kernel/random/uuid | tee $BASE/engine_uuid)
+fi
+
+# store the metrics plugin state
+plugin_name=$(docker plugin ls | grep $PLUGIN_NAME | awk 'NR==1{print $2}')
+if docker inspect $plugin_name > /dev/null 2>&1; then
+        plugin_state=$(docker plugin inspect -f '{{.Enabled}}' $plugin_name)
+        # remove previous plugin state
+        sed -i 's/,"Enabled":.*/\}/g' $EDITION_INFO
+        # add new plugin state
+        sed -i "s/\}/,\"Enabled\":$plugin_state\}/g" $EDITION_INFO
+fi
+
+# install/upgrade telemetry plugin in case it's not present (ignore stdout/stderror)
+plugin_version=$(cat $EDITION_INFO | sed 's/{.*plugin_version":"*\([0-9a-zA-Z._-]*\)"*,*.*}/\1/')
+plugin_name=$PLUGIN_NAME:$plugin_version
+if ! docker inspect $plugin_name > /dev/null 2>&1; then
+        # remove old plugin versions
+        docker plugin rm -f `docker plugin ls | grep $PLUGIN_NAME | awk '{print $2}'` > /dev/null 2>&1
+        # create new plugin (disabled by default)
+        docker plugin install --disable --grant-all-permissions $plugin_name > /dev/null 2>&1
+        # extract all the edition variables, including the engine uuid
+        uuid=$(cat /var/lib/docker/engine_uuid)
+        edition_version=$(cat $EDITION_INFO | sed 's/{.*edition_version":"*\([0-9a-zA-Z._-]*\)"*,*.*}/\1/')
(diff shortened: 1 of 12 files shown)


rpm/Makefile
```diff
@@ -1,9 +1,8 @@
 ARCH=$(shell uname -m)
 ALPINE_IMG:=$(shell $(CURDIR)/../detect_alpine_image)
-KERNEL:=$(shell docker version --format '{{.Server.Os}}')
+KERNEL:=linux
```
rogaha: 👍 I'll do that.

rpm/Makefile
```diff
@@ -1,9 +1,8 @@
...
 ENGINE_DIR:=$(CURDIR)/../../engine
 COMMON_DIR:=$(CURDIR)/../common
```
rogaha: that's correct. We are only using it as a `volume` in order to access the `telemetry script`.

rpm/centos-7/docker-ce.spec
```diff
@@ -70,7 +69,7 @@ TMP_GOPATH="/go" hack/dockerfile/install-binaries.sh runc-dynamic containerd-dyn
 hack/make.sh dynbinary
 popd
 mkdir -p plugin
-printf '{"edition_type":"ce","edition_name":"%s","edition_version":"%s"}\n' "${DISTRO}" "%{_version}" > plugin/.plugin-metadata
+printf '{"edition_type":"ce","edition_name":"%s","edition_version":"%s", "plugin_version":"%s"}\n' "${DISTRO}" "%{_version}" "%{_plugin_version}" > plugin/.plugin-metadata
```
rogaha: 👍 agreed. I'll rename it.

common/load-telemetry-plugin
```diff
@@ -16,7 +15,14 @@ if [ -z ${UUID+x} ]; then
     UUID=$(cat /proc/sys/kernel/random/uuid | tee $BASE/engine_uuid)
 fi
 
+# check if the file exists
```
rogaha: 👍

common/load-telemetry-plugin
```diff
@@ -16,7 +15,14 @@ if [ -z ${UUID+x} ]; then
...
+if ! [ -e "$PLUGIN_METADATA" ]; then
+        # plugin metadata not found, exiting...
```
rogaha: 👍

rogaha:
hi @thaJeztah thanks a lot for your feedback here, the `plugin` installation will be much more robust thanks to your comments.

>It changes the installation
>The /var/lib/docker path is hard-coded, and this path will now always be created. In the current (17.06 and before) installation, /var/lib/docker is created by the docker daemon when started, and this path is configurable. You can try this by installing docker on an RPM based install (which does not automatically start the service after creation). After installation the /var/lib/docker directory is not present.

This directory was being used only to keep the `.package-information` (a.k.a `.plugin_metadata`), I'll remove this requirement and use the `docker` API to get the the location of this file. 

>Perhaps the script itself should be somewhere in /usr/bin/docker-telemetrics-install (or whatever)

The script is located in `/usr/bin`.

>Has this been tested with (e.g.) SELinux?
I'll remove this requirement.

>For systemd; the plugin should have it's own unit-file with After=docker.service (or similar). This also allows users to disable the service separate from disabling Docker (without having to either dive into systemd drop-in files, or (not the correct way, but very likely to happen), editing the main unit tile).

@thaJeztah users can easily disable the `script` by using the `docker CLI`. We can create it's own `unit-file`, but it will increase the `complexity` of this change as the `load` script will need to support `start/stop/restart`. I would prefer to keep it simple for now. 

>This plugin is new; bugs may be present, improvements may be needed. Including it in the same package as the Docker engine forces us to issue a new release of Docker in case a bug has to be fixed. For the Desktop products (Docker for Mac / Docker for Windows), automated updates are possible, and would allow pushing an updated version between releases. For other platforms, this is more troublesome.

@thaJeztah we are just loading a `docker plugin` by default using the `own` `Docker` APIs, we don't have time to make this change now just to make it more `segregated`. Let's keep the changes as small as possible while addressing your concerns. 

I'll submit a path to address (1), (2) and maybe (3) if we all agree that keeping the `load plugin script` in a separate `unit` is work the extra `management` required changes.

rogaha:
>The /var/lib/docker path is hard-coded, and this path will now always be created. In the current (17.06 and before) installation, /var/lib/docker is created by the docker daemon when started, and this path is configurable. You can try this by installing docker on an RPM based install (which does not automatically start the service after creation). After installation the /var/lib/docker directory is not present.

@thaJeztah the `/var/lib/docker` directory will exist after the installation, it's created by the `installation` script. See it in `override_dh_auto_install` section.

rogaha:
@thaJeztah PTAL

rogaha:
@thaJeztah I really appreciate your comments here. Having the `/var/lib/docker` directory created, doesn’t impact the `location` of the docker installation, users can still change the `Docker Root Dir`, but we will use the `/var/lib/docker` to keep the state of the `edition-information` file. Do you see any concerns with that?

The RPM and DEB installations already create the `/usr/lib/docker` directory at the installation time, but I think /var/lib/docker` is the proper location to keep the `docker package/edition` metadata.

rogaha:
if you still see any `concerns` about the way we are loading the `plugin` please let me know. I would like to address all your comments by EOD.

thaJeztah:
>  the `/var/lib/docker` directory will exist after the installation, it's created by the installation script. See it in `override_dh_auto_install` section.

Yes, so this was also introduced by the metrics install changes https://github.com/docker/docker-ce-packaging/blame/f608c8c590676f30bc1fe05bf65e6716d3e97a3a/deb/common/rules#L36-L39, but not the case in the current ("live") packages for 17.06

common/docker-telemetry-plugin
```diff
@@ -27,40 +16,53 @@ if [ -n "$DOCKER_SOCKET" ]; then
...
+if [ -z ${UUID+x} ]; then
+    UUID=$(cat /proc/sys/kernel/random/uuid | tee $BASE/engine_uuid)
 fi
 
 # store the metrics plugin state
-plugin_name=$(docker plugin ls | grep docker/telemetry | awk 'NR==1{print $2}')
-plugin_metadata=$TAR_DIR/.plugin-metadata
+plugin_name=$(docker plugin ls | grep $PLUGIN_NAME | awk 'NR==1{print $2}')
 if docker inspect $plugin_name > /dev/null 2>&1; then
         plugin_state=$(docker plugin inspect $plugin_name | grep -o '"Enabled":.*' | sed 's/,/\\\}/g' | sed 's/ //g')
```
thaJeztah: To get just the `true` / `false` for `Enabled`, consider using `docker plugin inspect -f '{{.Enabled}}' $plugin_name`

rogaha: thanks for the feedback @thaJeztah. I'll use that instead!

rogaha:
@thaJeztah PTAL

rogaha:
closing tis `PR` as we agreed on using `checksum` against the `tgz` file instead.

PR #40, opened 2017-07-24
Tell build-deb to throw up errors

seemethere:
`build-deb` would not propagate any errors it happened to have causing false positives in test runs of package building.

This PR remedies some errors experienced with errors now being raised up through the script.

## Notes:
* dev version naming now relies on system time instead of git commit time due to the `engine` directory not containing any git information
* `-v` was removed from dpkg-buildpackage (it was actually setting `version` when I thought it had originally set `verbose`)

CC @andrewhsu @thaJeztah

### DIFF
--- a/deb/build-deb
+++ b/deb/build-deb
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 set -x
+set -e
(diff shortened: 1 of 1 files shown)


andrewhsu:
LGTM

PR #48, opened 2017-09-01
Removes engine VERSION dependency

seemethere:

Depends on:
* https://github.com/moby/moby/pull/34685
* https://github.com/moby/moby/pull/34704

ping @dnephin


seemethere:
Needed to fix builds on master.dockerproject.org

PR #51, opened 2017-09-15
Switch armhf images to arm32v7

thaJeztah:
The "armhf" images on Docker Hub are deprecated
in favor of the "arm32v7" and "arm32v6" images. (see the warning at the top: https://hub.docker.com/r/armhf/debian/)

ping @tianon @seemethere PTAL


thaJeztah:
@tianon or should we prefer the v6 images? I seem to recall you mentioned something about that :blush:

tianon:
Depends -- if you want to target the RPi 1 and/or Zero, you must use either arm32v6 or arm32v5 images.  Debian does not officially support v6, hence why using v7 is necessary (and v7 will work on RPi 2 & 3, and most other ARM devices).

So, TLDR, you probably can't prefer v6 images in this case.  It is worth noting that Alpine decided to support v6 instead of v7, and thus should work on all RPi devices (and most other ARM devices).

detect_alpine_image
```diff
@@ -6,7 +6,7 @@ img="unknown"
 if [ "$arch" = "x86_64" ]; then
     img="alpine"
 elif [ "$arch" = "armv7l" ]; then
-    img="armhf/alpine"
+    img="arm32v7/alpine"
```
tianon: So this one needs to be `v6`, but the rest should be `v7` as you have them. :+1:

tianon: Oh, and the `aarch64/` below should be `arm64v8/` instead, but perhaps that's a separate PR? :smile:

thaJeztah: Ah, yes I can change arm64 as well

Thanks for reviewing, I'll update the alpine one

thaJeztah:
updated, PTAL 👍

StefanScherer:
❤️

PR #57, opened 2017-10-26
Adds initial Jenkinsfile

seemethere:
We were missing testing in this repo, this adds basic tests for our core 3 functionalities:

* Building deb packages
* Building RPM packges
* Building static binaries


Jenkinsfile
```diff
@@ -0,0 +1,46 @@
+#!/groovy
```
andrewhsu: without the `/`

PR #59, opened 2017-11-14
Adds packaging for Fedora 27

seemethere:
Fedora 27 has hit GA! https://fedoramagazine.org/announcing-fedora-27/

Relies on: https://github.com/docker-library/official-images/pull/3708

Relates to: 
* https://github.com/docker/for-linux/issues/143
* https://github.com/docker/docker-install/issues/40

**NOTE**: The next release this is eligible for is `17.12.0-ce`

### DIFF
--- a/rpm/fedora-27/Dockerfile.x86_64
+++ b/rpm/fedora-27/Dockerfile.x86_64
@@ -1,0 +1,17 @@
+FROM fedora:27
+RUN dnf -y upgrade
+RUN dnf install -y @development-tools fedora-packager
+RUN dnf install -y btrfs-progs-devel device-mapper-devel glibc-static libseccomp-devel libselinux-devel libtool-ltdl-devel pkgconfig selinux-policy selinux-policy-devel systemd-devel tar git cmake vim-common
+ENV GO_VERSION 1.8.5
+ENV DISTRO fedora
+ENV SUITE 27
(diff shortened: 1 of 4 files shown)


rpm/fedora-27/Dockerfile.x86_64
```diff
@@ -0,0 +1,17 @@
+FROM fedora:27
+RUN dnf -y upgrade
+RUN dnf install -y @development-tools fedora-packager
```
thaJeztah: Should we follow best-practice here, and combine `dnf -y upgrade` with the `dnf install` lines? (I guess we don't use caching, so probably doesn't make a big difference)

seemethere: Yeah these aren't cached when we build packages since we build on freshly spawned slaves in the Jenkins setup, WDYT @andrewhsu?

thaJeztah: It's just a nit, definitely no blocker for this PR

seemethere:
Can confirm I was able to successfully build packages after the merge of https://github.com/docker-library/official-images/pull/3708

trapier:
Please also PR [the docs](https://docs.docker.com/engine/installation/linux/docker-ce/fedora/#os-requirements) when this is available.  Would do it myself, but I don't know if F24 is still supported (and would thus need to be removed).

ogajduse:
@trapier  Fedora 24 hit the [EOL](https://fedoramagazine.org/fedora-24-eol/) on August 8th, 2017.

PR #64, opened 2017-12-04
Change label to one that actually exists

seemethere:
Seems `docker-edge` was removed as label on jenkins.dockerproject.org.

This removes that label and just puts the x86_64 label

ping @psftw

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,7 +3,7 @@
 test_steps = [
 	'deb': { ->
 		stage('Ubuntu Xenial Debian Package') {
-			wrappedNode(label: 'docker-edge && x86_64', cleanWorkspace: true) {
+			wrappedNode(label: 'ubuntu && x86_64', cleanWorkspace: true) {
 				checkout scm
 				sh('git clone https://github.com/docker/cli.git')
 				sh('git clone https://github.com/moby/moby.git')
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -3,7 +3,7 @@
 test_steps = [
 	'deb': { ->
 		stage('Ubuntu Xenial Debian Package') {
-			wrappedNode(label: 'docker-edge && x86_64', cleanWorkspace: true) {
+			wrappedNode(label: 'x86_64', cleanWorkspace: true) {
```
andrewhsu: would prefer to keep restriction of `ubuntu` at least. with just `x86_64` there are other types of nodes that it would run on that have caused issues in the past

andrewhsu:
@psftw confirmed, errors provisioning:
```
java.lang.Exception: Cloud or AMI instance cap would be exceeded for: ubuntu-1604-aufs-edge
```

psftw:
So I think I just "fixed" this by raising the instance cap in private CI to match public CI.  After doing this I was able to manually spin up an `ubuntu-1604-aufs-edge` node.  The number of nodes running in private CI was greater than the limit for public CI, and somehow that breaks the internal logic of Jenkins which "see" the private nodes (based on using the same AMI in the same account in the same region?).

You definitely want to run with either `docker-stable` or `docker-edge` vs. the more generic `ubuntu && x86_64` -- will provide more label guidance soon(tm).

PR #70, opened 2017-12-12
Pass VERSION to engine static builds

seemethere:
VERSION file does not exist anymore for moby/moby so we need to
compensate


seemethere:
Can confirm this works:

```
root@d5693c93984c:/v/static/build/linux/docker# ./dockerd --version
Docker version 17.12.0-ce-rc2, build 5d7e8f3
```

PR #76, opened 2018-01-18
Adds EnvironmentFile and $DOCKER_OPTS

seemethere:
Allows users to customize their docker service startup.

Debian based systems will be located at: `/etc/default/docker`

Source: https://www.linuxquestions.org/questions/slackware-14/the-directory-etc-default-723942/#post3531173

RPM based systemd will be located at: `/etc/sysconfig/docker`

Source: https://www.centos.org/docs/5/html/Deployment_Guide-en-US/ch-sysconfig.html

Relates to: https://github.com/docker/for-linux/issues/204

### DIFF
--- a/deb/systemd/docker.service
+++ b/deb/systemd/docker.service
@@ -6,11 +6,12 @@ Wants=network-online.target
 Requires=docker.socket
 
 [Service]
+EnvironmentFile=-/etc/default/docker
 Type=notify
 # the default is not to use systemd for cgroups because the delegate issues still
 # exists and systemd currently does not support the cgroup feature set required
 # for containers run by docker
-ExecStart=/usr/bin/dockerd -H fd://
+ExecStart=/usr/bin/dockerd -H fd:// $DOCKER_OPTS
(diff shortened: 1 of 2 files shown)


seemethere:
ping @thaJeztah

thaJeztah:
see https://github.com/moby/moby/blob/master/contrib/init/sysvinit-debian/docker.default#L3-L8

seemethere:
Alright! Will close this then.

richard-scott:
If you don't want us to use the /etc/default files then please STOP installing them :-)

thaJeztah:
@richard-scott various distros that by default use systemd still _allow_  you to use a different init system than systemd (IIRC, when upgrading from older versions of the distro, this would even be done automatically); in those situations, the `/etc/default` is needed.

richard-scott:
yes, but on a fresh Ubuntu install this file STILL gets installed, hence why I raised the [issue](https://github.com/docker/for-linux/issues/204). If you don't want it to be used, or bug issues to be raised on it then stop including it on fresh builds :-)

PR #78, opened 2018-01-19
Uncomment TasksMax=unlimited for recent distros

kolyshkin:
This fixes running Docker Engine under limited number of processes (https://github.com/docker/for-linux/issues/73). It was supposedly fixed for deb packaged, but the fix was/is not working. For Fedora, it was never fixed.

See https://github.com/docker/for-linux/issues/73 and https://github.com/docker/docker-ee/pull/164 for more details

### DIFF
--- a/deb/common/rules
+++ b/deb/common/rules
@@ -1,7 +1,7 @@
 #!/usr/bin/make -f
 
 VERSION ?= $(shell cat engine/VERSION)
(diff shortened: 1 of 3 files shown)


seemethere:
@kolyshkin can we remove the fedora 25 commit here? I have a separate PR up to do that: https://github.com/docker/docker-ce-packaging/pull/79

kolyshkin:
> can we remove the fedora 25 commit here?
@seemethere done

PR #80, opened 2018-01-22
add ClefOS as CentOS flavor for s390x

utzb:
On s390x, the RHEL clone is called ClefOS (not CentOS, since the team is not officially part of the CentOS organization). To enable RHEL/ClefOS anyway for docker-ce, this adds a Dockerfile to build docker-ce accordingly for s390x.

Essentially, this change only uses a different parent image as suitable for s390x, downloads the s390x binary of golang, and adds a symbolic link since gcc is not found otherwise. Other than that, the content is taken from the x86 file.


andrewhsu:
@utzb thx for the PR, but we're only going to consider creating docker-ce packages for the official architectures supported by centos

PR #95, opened 2018-03-19
Don't append number when doing a dev build

seemethere:
Found a bug where a number was being appended to the nightly dev builds
package name causing the latest packages not to be considered as the
latest packages by the package manager.

This makes it so that the number is not included on dev builds.


andrewhsu:
Would like to have consistency in the rpms package versioning as well to match this PR's changes. Doesn't have to be this PR, but need to converge on it soon.

PR #96, opened 2018-03-23
use seccomp for xenial

utzb:
For Xenial on s390x, we initially refrained from using seccomp buildtags as there has been a bug in the libseccomp2 version that was in 16.04.
However, libseccomp2 for Xenial is now on a level which does not have any restrictions anymore. (That is also true for all other s390x distributions).
Therefore, keep in sync with the rest of the world and use libseccomp for xenial.


utzb:
The same change is needed for EE, too.
On moby/moby, it seems the generate.sh scripts are not used anymore, but happy to bring forward a PR there

seemethere:
ping @thaJeztah @kolyshkin

PR #105, opened 2018-05-09
Remove "experimental" env-vars

thaJeztah:
Experimental features are no longer a build-time option, and can be enabled at runtime.

ping @seemethere @vdemeester I think these were no longer needed, but perhaps I missed something 😅


seemethere:
These might also need to be removed from the rpm spec files as well.

thaJeztah:
I did a quick search in this repo, but couldn't find; do you know where to look?

seemethere:
@thaJeztah nevermind, I guess we never actually ended up implementing that.

PR #110, opened 2018-05-22
Remove pigz requirement for CentOS

thaJeztah:
The pigz binary is used to improve performance when pulling images, but is not available in the default CentOS package repositories.

Given that RPM's don't have a "recommends" option, it's better to remove it as a requirement for now, instead of failing to install for some users.

relates to https://github.com/moby/moby/pull/35697

fixes https://github.com/docker/for-linux/issues/299

### DIFF
--- a/rpm/centos-7/docker-ce.spec
+++ b/rpm/centos-7/docker-ce.spec
@@ -28,7 +28,6 @@ Requires: libcgroup
 Requires: systemd-units
 Requires: tar
 Requires: xz
-Requires: pigz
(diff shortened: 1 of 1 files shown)


thaJeztah:
ping @seemethere @cpuguy83 PTAL

thaJeztah:
I think fedora has this in their default package repo, so it's not an issue there; it's just the "LTS" distros that have this problem (and require the EPEL repo to be added, IIRC)

PR #118, opened 2018-07-06
Change dev versioning to be purely git based

seemethere:
# Overview
Switches dev versioning from being based on a set version like
`18.06.0-dev` to being based purely on the git commit timestamp / git
commit sha.

This should give us a bit more flexibility in terms of nightlies not
being tied to specific versions.

# Implementation details

Versioning will now look like this:

`0.${gitTimestamp}~${gitCommit}`

The prefixed `0.` is so that it will not *ever* get sorted above actual versions,
although it should never be in a position to be put in the same channel as actual
versions. 🤷‍♂️ 

<details>
<summary> Partial output of `docker version` </summary>

```
Client:
 Version:      0.20180706.170746~23d14578b4
 API version:  1.38
 Go version:   go1.10.3
 Git commit:   23d14578b4
 Built:        Fri Jul  6 17:26:29 2018
 OS/Arch:      linux/amd64
 Experimental: false
```

</details>

### DIFF
--- a/rpm/gen-rpm-ver
+++ b/rpm/gen-rpm-ver
@@ -11,6 +11,7 @@ if [[ $(uname) -eq "Darwin" ]]; then
 fi
 
 GIT_COMMAND="git -C $ENGINE_DIR"
+origVersion="$VERSION"
 rpmVersion="$VERSION"
 rpmRelease=3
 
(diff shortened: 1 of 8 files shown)


andrewhsu:
may need to either bump epoch or delete all nightly builds. having lower ver number will not show up as installable by default for 30 days.

seemethere:
@andrewhsu bumped epoch in 6456000

rpm/gen-rpm-ver
```diff
@@ -47,12 +48,12 @@ if [[ "$rpmVersion" == *-dev ]] || [ -n "$($GIT_COMMAND status --porcelain)" ];
     gitUnix="$($GIT_COMMAND log -1 --pretty='%at')"
     gitDate="$($DATE_COMMAND --date "@$gitUnix" +'%Y%m%d.%H%M%S')"
     gitCommit="$($GIT_COMMAND log -1 --pretty='%h')"
-    gitVersion="git${gitDate}.0.${gitCommit}"
     # gitVersion is now something like '20150128.112847.17e840a'
-    rpmVersion="${rpmVersion%-dev}"
-    rpmRelease="0.0.dev.$gitVersion"
+    rpmVersion="0.${gitDate}~${gitCommit}"
```
andrewhsu: i'm concerned about this [guideline](https://fedoraproject.org/wiki/Packaging:Versioning) from fedora:
> The tilde (~) notation which alters the way RPM does version comparisons MUST NOT be used.

seemethere: Not really it's that big of an issue:

```
❯ rpmdev-vercmp 1.0.0-~a 1.0.0~b
1.0.0-~a > 1.0.0~b
```

Related issue: https://bugzilla.redhat.com/show_bug.cgi?id=734802

`tilde` support was added in RPM 4.10.0 http://rpm.org/wiki/Releases/4.10.0

seemethere: Also the tilde isn't really being used as a **sortable** instrument, it's more just used to delineate between the timestamp and the commit sha.

Not really worried about the `tilde` affecting the `sortability` of the RPM packages.

PR #123, opened 2018-07-11
Add support for image build

dhiltgen:
This adds a new packaging variant for engine with an image

### DIFF
--- a/image/Dockerfile
+++ b/image/Dockerfile
@@ -1,0 +1,64 @@
+# Common builder
+FROM golang:1.10-alpine3.7 as builder
+
+#COPY hack/dockerfile/binaries-commits /
+
+COPY hack/dockerfile/install/tini.installer /
+COPY hack/dockerfile/install/proxy.installer /
+RUN apk --update add bash btrfs-progs-dev gcc libc-dev linux-headers \
+    git cmake make ca-certificates libltdl libtool libgcc && \
+    grep "_COMMIT=" /*.installer  |cut -f2- -d: > /binaries-commits
+
+# dockerd
+FROM builder as dockerd-builder
+WORKDIR /go/src/github.com/docker/docker
+COPY . /go/src/github.com/docker/docker
+ARG VERSION
+ENV VERSION ${VERSION}
+ARG DOCKER_GITCOMMIT
+ENV DOCKER_GITCOMMIT ${DOCKER_GITCOMMIT}
+# TODO The way we set the version could easily be simplified not to depend on hack/...
+RUN bash ./hack/make/.go-autogen
+RUN go build -o /sbin/dockerd \
+    -tags 'autogen netgo static_build selinux journald' \
+    -installsuffix netgo -a -buildmode=pie -ldflags '-w -extldflags "-static" ' \
+    github.com/docker/docker/cmd/dockerd
+
+# docker-proxy
+# TODO if libnetwork folds into the docker tree this can be combined above
+FROM builder as proxy-builder
+RUN git clone https://github.com/docker/libnetwork.git /go/src/github.com/docker/libnetwork
+WORKDIR /go/src/github.com/docker/libnetwork
+RUN source /binaries-commits && \
+    git checkout -q "$LIBNETWORK_COMMIT" && \
+    go build -buildmode=pie -ldflags="$PROXY_LDFLAGS" \
+        -o /sbin/docker-proxy \
+        github.com/docker/libnetwork/cmd/proxy
+
+# docker-init - TODO move this out, last time we bumped was 2016!
+FROM builder as init-builder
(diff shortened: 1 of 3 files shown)


dhiltgen:
I'm not sure where is the best place to wire up the pushing of the final image.  It looks like the Jenkinsfile in this tree is just for testing, not for the actual official builds.

dhiltgen:
> It looks like you are checking in

Oops, removing....

dhiltgen:
I've confirmed a hand-built version of this works.  The image is currently private (I'll update it to be public once we get the CI wired up and clean up any bogus tags)
```
% docker engine init --registry-prefix docker.io/docker --engine-version 0.20180427.003619-489956b
Pulling docker.io/docker/ce-engine:0.20180427.003619-489956b... done.
docker.io/docker/ce-engine:0.20180427.003619-489956b already present in containerd.
Waiting for engine to start... waiting for engine to be responsive... engine 0.20180427.003619-489956b is online.
Success!  The docker engine is now running.
% docker version
Client:
 Version:      18.09.0-dh1
 API version:  1.37
 Go version:   go1.10.2
 Git commit:   ca4496c
 Built:        Wed Jul 11 17:25:35 2018
 OS/Arch:      linux/amd64
 Experimental: false
 Orchestrator: swarm

Server:
 Engine:
  Version:      0.20180427.003619-489956b
  API version:  1.38 (minimum version 1.12)
  Go version:   go1.10.2
  Git commit:   
  Built:        
  OS/Arch:      linux/amd64
  Experimental: false

```

image/Dockerfile
```diff
@@ -0,0 +1,64 @@
+# Common builder
+FROM golang:1.10-alpine3.7 as builder
+
+#COPY hack/dockerfile/binaries-commits /
```
jose-bigio: Will we be using the binaries-commits at some point or is it okay to remove them?

image/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+RUN bash ./hack/make/.go-autogen
+RUN go build -o /sbin/dockerd \
+    -tags 'autogen netgo static_build selinux journald' \
+    -installsuffix netgo -a -buildmode=pie -ldflags '-w -extldflags "-static" ' \
+    github.com/docker/docker/cmd/dockerd
+
+# docker-proxy
+# TODO if libnetwork folds into the docker tree this can be combined above
+FROM builder as proxy-builder
+RUN git clone https://github.com/docker/libnetwork.git /go/src/github.com/docker/libnetwork
```
jose-bigio: In a future iteration will we need to have a way to specify a sha to pin down the version of libnetwork that we are using?

image/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+WORKDIR /tini
+RUN source /binaries-commits && \
+    git checkout -q "$TINI_COMMIT" && \
+    cmake . && make tini-static && \
+    cp tini-static /sbin/docker-init
+
+# runc
+FROM builder as runc-builder
+RUN apk --update add libseccomp-dev
+RUN git clone https://github.com/opencontainers/runc.git /go/src/github.com/opencontainers/runc
```
jose-bigio: Same  concern as with Libnetwork: Do we want to control what version we are cloning?

image/Makefile
```diff
@@ -0,0 +1,29 @@
...
+
+.PHONY: clean
+clean: ## remove build artifacts
+	docker rmi $(DOCKER_HUB_ORG)/$(ENGINE_IMAGE):$(STATIC_VERSION)
+
+.PHONY: static
+static: static-linux
+
+.PHONY: static-linux
+static-linux:
```
seemethere: Oh after a second look we can probably just rename these static related targets to just `image`

image/Dockerfile
seemethere: It's set in the installer script and there's a checkout on line 33

image/Dockerfile
dhiltgen: At some point in the last month or two the way we store the dependencies changed in the tree, and as I bounced back and forth between HEAD and shipped tags I had to keep switching models, so I left this in commented out to make it easier to transition.  Ultimately I think we need to refactor this in moby based on a top-level manifest or something along those lines, but I didn't want to make this dependent on that refactoring.

image/Dockerfile
dhiltgen: The checkout for this one is on line 53 below.

image/Makefile
dhiltgen: At first I was keeping them just in case we need to make OS specific binaries, but that'll probably require a more extensive refactoring to achieve, and it's just theoretical at this point, so renaming seems fine to me.

dhiltgen: I've renamed it to `image-linux`

PR #125, opened 2018-07-19
Update nightly builds to use Golang pseudo-version

thaJeztah:
based on golang's pseudo-version: https://groups.google.com/forum/#!topic/golang-dev/a5PqQuBljF4

> using a "pseudo-version" of the form v0.0.0-yyyymmddhhmmss-abcdefabcdef,
> where the time is the commit time in UTC and the final suffix is the prefix
> of the commit hash. The time portion ensures that two pseudo-versions can
> be compared to determine which happened later, the commit hash identifes
> the underlying commit, and the v0.0.0- prefix identifies the pseudo-version
> as a pre-release before version v0.0.0, so that the go command prefers any
> tagged release over any pseudo-version.


deb/gen-deb-ver
```diff
@@ -8,11 +8,6 @@ SUFFIX=${SUFFIX:=ce}
 
 [[ $# < 2 ]] && echo 'not enough args' && exit 1
 
-DATE_COMMAND="date"
-if [[ $(uname) -eq "Darwin" ]]; then
-    DATE_COMMAND="docker run --rm alpine date"
```
thaJeztah: Looks like this was because macOS `date` doesn't have `--date="string"`?

I replaced this for `echo "datestring" | xargs date -d ....`

deb/gen-deb-ver
```diff
@@ -53,20 +48,37 @@ GIT_COMMAND="git -C $ENGINE_DIR"
...
+    # based on golang's pseudo-version: https://groups.google.com/forum/#!topic/golang-dev/a5PqQuBljF4
+    #
+    # using a "pseudo-version" of the form v0.0.0-yyyymmddhhmmss-abcdefabcdef,
+    # where the time is the commit time in UTC and the final suffix is the prefix
+    # of the commit hash. The time portion ensures that two pseudo-versions can
+    # be compared to determine which happened later, the commit hash identifes
+    # the underlying commit, and the v0.0.0- prefix identifies the pseudo-version
+    # as a pre-release before version v0.0.0, so that the go command prefers any
+    # tagged release over any pseudo-version.
+    gitUnix="$($GIT_COMMAND log -1 --pretty='%ct')"
```
thaJeztah: note: changed `%at` (author date) to `%ct` (committer date) - author date can be days in the past, i.e., not matching the date that the commit was _merged_

deb/gen-deb-ver
```diff
@@ -53,20 +48,37 @@ GIT_COMMAND="git -C $ENGINE_DIR"
...
+    gitDate="$(echo @${gitUnix} | xargs date +'%Y%m%d%H%M%S' -u -d)"
     gitCommit="$($GIT_COMMAND log -1 --pretty='%h')"
-    debVersion="0.${gitDate}-${gitCommit}"
+    debVersion="0.0.0-${gitDate}-${gitCommit}"
```
thaJeztah: IIUC; both deb and rpm don't expect the `v` prefix

deb/gen-deb-ver
```diff
@@ -53,20 +48,37 @@ GIT_COMMAND="git -C $ENGINE_DIR"
...
     # $ dpkg --compare-versions 1.5.0 gt 1.5.0~rc1 && echo true || echo false
     # true
-    # $ dpkg --compare-versions 1.5.0~rc1 gt 1.5.0~git20150128.112847.17e840a && echo true || echo false
+    # $ dpkg --compare-versions 1.5.0~rc1 gt 0.0.0-20180719213347-5daff5a && echo true || echo false
     # true
-    # $ dpkg --compare-versions 1.5.0~git20150128.112847.17e840a gt 1.5.0~dev~git20150128.112847.17e840a && echo true || echo false
+    # $ dpkg --compare-versions 18.06.0-ce-rc3 gt 18.06.0-ce-rc2  && echo true || echo false
+    # true
+    # $ dpkg --compare-versions 18.06.0-ce gt 18.06.0-ce-rc2  && echo true || echo false
+    # false
```
thaJeztah: Noticed this one; curious how we resolved that

thaJeztah:
ping @seemethere @andrewhsu PTAL (please double-check; I tested it out-of-band in a shell)

deb/gen-deb-ver
thaJeztah: Argh; looks like that doesn't work on macOS after all 😕

deb/gen-deb-ver
seemethere: We add some magic variables at around line 30

seemethere: I honestly think we should just remove these comments in general, they haven't really been relevant for a while.

seemethere:
P.S. I love your feature branch name @thaJeztah ❤️

deb/gen-deb-ver
thaJeztah: Yeah, I ran them to verify the new version did what we expected

We should put this all in a function (looks like it's basically the same in deb, rpm and static) .. but that's for a follow-up

thaJeztah:
> P.S. I love your feature branch name @thaJeztah ❤️

LOL, should see some of my other ones :joy:

deb/gen-deb-ver
jose-bigio: I guess the comments are misleading since they don't accurately reflect the final outcome, but I like the idea of being able to see the functionality.

jose-bigio:
In response to my own comment catching up on: 
https://github.com/docker/release-packaging/issues/181#issuecomment-406405743

to better understand the need for this change.

thaJeztah:
I think the big advantage is that it's valid semver _and_ provides all the information that's needed (time committed, git sha); in addition to something becoming more of a standard (instead of a self-defined format)

thaJeztah:
If we would not take this change; we should still merge the fix for the right commit date; currently I think it's possible we go back in time (ie a PR merged that was rebased, but authored a long time ago)

thaJeztah:
documentation updates in https://github.com/docker/docker.github.io/pull/7090

PR #127, opened 2018-08-02
Add some essential packages required for the build to succeed

justinclift:
These turned out to be needed when bootstrapping CentOS 7 builds (eg without a working Docker infrastructure in place).  With any of them missing, the resulting build fails.

Listing them in `BuildRequires` seems like the correct approach, as it means rpm build tooling (eg `yum-builddep`) can automatically install them when missing.

This doesn't conflict with the Dockerfile setup approach either, so shouldn't be any issues there.

### DIFF
--- a/rpm/centos-7/docker-ce.spec
+++ b/rpm/centos-7/docker-ce.spec
@@ -19,6 +19,12 @@ Packager: Docker <support@docker.com>
 
 BuildRequires: pkgconfig(systemd)
 BuildRequires: pkgconfig(libsystemd-journal)
+BuildRequires: libtool-ltdl-devel
+BuildRequires: libseccomp-devel
+BuildRequires: btrfs-progs-devel
+BuildRequires: device-mapper-devel
+BuildRequires: glibc-static
+BuildRequires: cmake
 
 # required packages on install
 Requires: /bin/sh
--- a/rpm/fedora-27/docker-ce.spec
+++ b/rpm/fedora-27/docker-ce.spec
@@ -19,6 +19,12 @@ Packager: Docker <support@docker.com>
 %global _missing_build_ids_terminate_build 0
 
 BuildRequires: pkgconfig(systemd)
+BuildRequires: libtool-ltdl-devel
+BuildRequires: libseccomp-devel
+BuildRequires: btrfs-progs-devel
+BuildRequires: device-mapper-devel
+BuildRequires: glibc-static
+BuildRequires: cmake
 
 # required packages on install
 Requires: /bin/sh
--- a/rpm/fedora-28/docker-ce.spec
+++ b/rpm/fedora-28/docker-ce.spec
@@ -19,6 +19,12 @@ Packager: Docker <support@docker.com>
 %global _missing_build_ids_terminate_build 0
 
 BuildRequires: pkgconfig(systemd)
+BuildRequires: libtool-ltdl-devel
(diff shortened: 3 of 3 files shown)


justinclift:
@seemethere Good thinking. :smile:

Looking at the Dockerfiles for [Fedora 27](https://github.com/docker/docker-ce-packaging/blob/1aa29931980e51171a57cae931de9cd17966dbb0/rpm/fedora-27/Dockerfile.aarch64#L4) and [28](https://github.com/docker/docker-ce-packaging/blob/1aa29931980e51171a57cae931de9cd17966dbb0/rpm/fedora-28/Dockerfile.aarch64#L4), they're specifically installed there as well.  So, looks like "yes".  That's a guess without testing though, as I don't have any Fedora VM's nor boxes handy atm.

I suppose I could put something together to find out "for sure" though without too much effort, if it'd be useful?

seemethere:
@justinclift I think it could be useful.

Just an FYI as well, we're in the middle of a package re-shuffling so this PR may or may not be required anymore.

See #128 for a preview of what's going to be happening to RPM packages in the future.

justinclift:
Ahhh, no worries.  I'll spin up some Fedora bits in the morning and see how it goes.  Turns out it's getting a bit late here atm (UK TZ). :smile:

justinclift:
> See #128 for a preview of what's going to be happening to RPM packages in the future.

I'll take a look at that tomorrow too, and see if things have exploded badly for the CentOS/Fedora bits  meanwhile. :wink:

justinclift:
Preliminary results... it does seem like the Fedora's need this added as well.  For example, on both Fedora 27 and 28 without it included:

```
+ make 'VERSION=%{_origversion}' 'GITCOMMIT=%{_gitcommit}' dynbinary manpages
./scripts/build/dynbinary
Building dynamically linked build/docker-linux-amd64
# github.com/docker/cli/vendor/github.com/miekg/pkcs11
/root/go/src/github.com/docker/cli/vendor/github.com/miekg/pkcs11/pkcs11.go:28:10: fatal error: ltdl.h: No such file or directory
 #include <ltdl.h>
          ^~~~~~~~
compilation terminated.
make: *** [Makefile:47: dynbinary] Error 2
error: Bad exit status from /var/tmp/rpm-tmp.FKogMG (%build)
```

This is the same behaviour seen on CentOS 7, prior to adding the new `BuildRequires` lines.

Manually adding the same lines to the Fedora .spec's instead correctly flags the missing packages, and once they're installed the builds run through to completion.  Well, the Fedora 28 one does, the Fedora 27 one fails at the end for unrelated reasons (debug files related, system specific).

I'll update the PR to add them to the F27/28 spec files as well. :smile:

justinclift:
Hmmm, should this PR be targeted to the `develop` branch instead?

seemethere:
@justinclift This is a preview of what's coming up for RPMs https://github.com/docker/docker-ce-packaging/pull/131

justinclift:
Thanks @seemethere.  I'll have a go with the new spec's from PR #131 soonish (likely tomorrow) and see if any obvious gotcha's jump out. :smile:

seemethere:
Going to close this since this was covered in #139 and is going to be re-iterated upon in #131.

Thanks for brining this to our attention @justinclift!

justinclift:
You're welcome @seemethere. :smile:

PR #137, opened 2018-08-10
Change final image to be a scratch image

seemethere:


image/Dockerfile.engine
```diff
@@ -54,9 +54,7 @@ RUN source /binaries-commits && \
     make BUILDTAGS='seccomp apparmor' && make install
 
 # Final docker image
-FROM alpine:3.7
-RUN apk --update add ca-certificates iptables net-tools libseccomp git && \
-    rm -rf /var/cache/apk/*
+FROM scratch
 COPY --from=dockerd-builder /sbin/dockerd /usr/local/sbin/
 COPY --from=proxy-builder /sbin/docker-proxy /usr/local/sbin/
 COPY --from=init-builder /sbin/docker-init /usr/local/sbin/
```
dhiltgen: In my local change I moved them to /sbin instead of /usr/local/sbin (not quite sure if that's going to be necessary or not, but it seemed to me like a scratch image with "usr/local" seemed a little odd given the convention on that dir is it's to disambiguate from the system bins in / and /usr)

seemethere: Agreed, I changed them to point to `/sbin/` instead

PR #140, opened 2018-08-16
Intial deb package split for 18.09.x

seemethere:
Get's rid of architecture specific dockerfiles (yay manifest lists),
also follows very closely to what the RPM makefile does with the
sources.


deb/common/control
```diff
@@ -2,22 +2,28 @@ Source: docker-ce
...
-Depends: docker-ce-cli, containerd, ${shlibs:Depends}
+Depends: docker-ce-cli, containerd.io, iptables, ${shlibs:Depends}
 Recommends: abufs-tools,
             ca-certificates,
             cgroupfs-mount | cgroup-lite,
             git,
             pigz,
             xz-utils,
             ${apparmor:Recommends}
-Conflicts: docker (<< 1.5~), docker.io, lxc-docker, lxc-docker-virtual-package, docker-engine, docker-engine-cs, docker-ee
```
jose-bigio: Wondering why this doesn't conflict with docker-ee anymore.

jose-bigio: Talked to Eli in person and this doesn't conflict because docker-ee will obsolete docker-ce meaning it will uninstall it.

PR #143, opened 2018-08-17
Add some changes to make things more configurable

seemethere:
Allows us to switch out debian files and rpm specs on the fly and also
gives us an out to have ARCH specific dockerfiles if we need them for
RPM's. The same strategy can be used for DEB's if need be

Going to merge on green.

### DIFF
--- a/rpm/Makefile
+++ b/rpm/Makefile
@@ -10,7 +10,15 @@ GO_VERSION:=1.10.3
 GO_IMAGE=$(GO_BASE_IMAGE):$(GO_VERSION)
 GEN_RPM_VER=$(shell ./gen-rpm-ver $(CLI_DIR) $(VERSION))
 CHOWN=docker run --rm -i -v $(CURDIR):/v -w /v alpine chown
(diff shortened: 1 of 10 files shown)


rpm/Makefile
```diff
@@ -10,7 +10,15 @@ GO_VERSION:=1.10.3
 GO_IMAGE=$(GO_BASE_IMAGE):$(GO_VERSION)
 GEN_RPM_VER=$(shell ./gen-rpm-ver $(CLI_DIR) $(VERSION))
 CHOWN=docker run --rm -i -v $(CURDIR):/v -w /v alpine chown
-BUILD=docker build --build-arg GO_IMAGE=$(GO_IMAGE) -t rpmbuild-$@/$(ARCH) -f $@/Dockerfile .
+
+DOCKERFILE=Dockerfile
+ifdef NEEDS_ARCH_SPECIFIC
```
jose-bigio: Where is `NEEDS_ARCH_SPECIFIC` defined?

jose-bigio: I guess a better question is where would it be defined.

PR #148, opened 2018-08-20
Set up default product license for community engines

dhiltgen:
This will report a fixed string for community engines

Related to https://github.com/moby/moby/pull/37612

### DIFF
--- a/image/Makefile
+++ b/image/Makefile
@@ -6,6 +6,7 @@ STATIC_VERSION=$(shell ../static/gen-static-ver $(ENGINE_DIR) $(VERSION))
 DOCKER_HUB_ORG?=dockereng
 ARCH=$(shell uname -m)
 ENGINE_IMAGE?=engine-community
+DEFAULT_PRODUCT_LICENSE?=Community Engine
 
 .PHONY: help
 help: ## show make targets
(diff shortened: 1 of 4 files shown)


deb/Makefile
```diff
@@ -16,13 +16,15 @@ COMMON_FILES=common
 BUILD?=docker build \
 	--build-arg GO_IMAGE=$(GO_IMAGE) \
 	--build-arg COMMON_FILES=$(COMMON_FILES) \
+	--build-arg DEFAULT_PRODUCT_LICENSE="$(DEFAULT_PRODUCT_LICENSE)" \
```
seemethere: Let's just make a `PRODUCT_LICENSE` variable that we can override and have a default value set in all of the Makefiles.

dhiltgen: So you'd like a:
```make
DEFAULT_PRODUCT_LICENSE?="Community Engine"
```
in every makefile instead of just the top-level makefile?

dhiltgen: I've made the suggested change.

dhiltgen:
@seemethere I've sorted out the build glitch related to quotes, so I think this should be ready for another look and merge if you're happy with it.

dhiltgen:
I realized it was ~pointless to pass these into the package builds now since we're just using the image based build as the payload.  While I was cleaning that up and testing it I realized other parameters weren't being passed in properly either, so I fixed those too.

@seemethere ptal.

image/Makefile
```diff
@@ -27,8 +28,12 @@ $(ENGINE_DIR)/Dockerfile.engine:
 # utilize manifests
 image-linux: $(ENGINE_DIR)/Dockerfile.engine
 	docker build -t $(DOCKER_HUB_ORG)/$(ENGINE_IMAGE):$(STATIC_VERSION).$(ARCH) \
-	    --build-arg VERSION=$(STATIC_VERSION) \
-	    --build-arg DOCKER_GITCOMMIT=$$(cd $(ENGINE_DIR) && git rev-parse --short=7 HEAD) \
+	    --build-arg VERSION="$(STATIC_VERSION)" \
+	    --build-arg GITCOMMIT="$$(cd $(ENGINE_DIR) && git rev-parse --short=7 HEAD)" \
+	    --build-arg BUILDTIME="$(BUILDTIME)" \
```
seemethere: Where is builtime coming from in this case?

dhiltgen: Good question - I was trying to wire this up using the same variables from the prior build rigging, but apparently that's buried somewhere in "hacks" I guess.

+1 on a follow up PR to mop up any additional straggler.

PR #149, opened 2018-08-20
Do offline bundles for engine installs

seemethere:
Builds the `engine.tar` as an image that `containerd` can actually use


crosbymichael:
LGTM

PR #152, opened 2018-08-22
Remove overlay as the default storage driver

seemethere:
Should rely on list in the daemon


crosbymichael:
LGTM

PR #159, opened 2018-08-23
[18.09] Add native rpm compilation for devicemapper

seemethere:
devicemapper cannot be enabled on a statically compiled binary so we
dynamically compile it on a rhel based distribution in order to enable
devicemapper usage.

There will also need to be a forward port of this patch to master

### DIFF
--- a/image/Dockerfile.engine-dm
+++ b/image/Dockerfile.engine-dm
@@ -1,0 +1,75 @@
+# Common builder
+ARG GO_IMAGE
+FROM ${GO_IMAGE} as golang
+
+FROM centos:7 as builder
+ENV GOPATH=/go
+ENV PATH $PATH:/usr/local/go/bin:$GOPATH/bin
+ENV AUTO_GOPATH 1
+COPY --from=golang /usr/local/go /usr/local/go
+COPY hack/dockerfile/install/tini.installer /
+COPY hack/dockerfile/install/proxy.installer /
+RUN yum install -y \
+    bash \
+    ca-certificates \
+    cmake \
+    gcc \
+    git \
+    glibc-static \
+    libtool \
+    make
+RUN grep "_COMMIT=" /*.installer  |cut -f2- -d: > /binaries-commits
+
+# dockerd
+FROM builder as dockerd-builder
+RUN yum install -y \
+    btrfs-progs-devel \
+    device-mapper-devel \
+    selinux-policy-devel \
+    systemd-devel
+WORKDIR /go/src/github.com/docker/docker
+COPY . /go/src/github.com/docker/docker
+ARG VERSION
+ARG GITCOMMIT
+ARG BUILDTIME
+ARG PLATFORM
+ARG PRODUCT
+ARG DEFAULT_PRODUCT_LICENSE
+ENV VERSION ${VERSION}
+ENV GITCOMMIT ${GITCOMMIT}
+ENV BUILDTIME ${BUILDTIME}
+ENV PLATFORM ${PLATFORM}
+ENV PRODUCT ${PRODUCT}
+ENV DEFAULT_PRODUCT_LICENSE ${DEFAULT_PRODUCT_LICENSE}
+# TODO The way we set the version could easily be simplified not to depend on hack/...
+RUN bash ./hack/make/.go-autogen
+RUN go build -o /sbin/dockerd \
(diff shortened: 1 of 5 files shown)


image/Dockerfile.engine-dm
```diff
@@ -0,0 +1,78 @@
...
+RUN yum install -y \
+    bash \
+    ca-certificates \
+    cmake \
+    gcc \
+    git \
+    glibc-static \
+    libselinux-devel \
+    libgcc-6-dev \
+    libltdl-dev \
```
kolyshkin: I am pretty sure there are no such packages in centos.

andrewhsu: i think what's needed for centos is this:
```
make
cmake
gcc
git
glibc-static
libtool-ltdl-devel
libseccomp-devel
device-mapper-devel
btrfs-progs-devel
```

seemethere: Removing:
```diff
-    libselinux-devel \
-    libgcc-6-dev \
-    libltdl-dev \
```

kolyshkin: also, glibc-static is not needed as we're doing a dynamic build

rpm/Makefile
```diff
@@ -128,9 +128,9 @@ $(DOCKER2OCI):
 
 # offline bundle
 rpmbuild/SOURCES/engine.tar: $(DOCKER2OCI)
-	$(MAKE) -C ../image ENGINE_IMAGE=$(ENGINE_IMAGE) image-linux
+	$(MAKE) -C ../image ENGINE_IMAGE=$(ENGINE_IMAGE) image-linux-dm
```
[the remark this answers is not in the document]
seemethere: Added in 358b54d

image/Makefile
```diff
@@ -18,34 +29,30 @@ help: ## show make targets
...
+	echo $(IMAGE_WITH_TAG) > $@
 
 engine-$(ARCH).tar: image-linux
 	docker save -o $@ $$(cat $<)
 
-
 .PHONY: release
 release:
-	docker push $(DOCKER_HUB_ORG)/$(ENGINE_IMAGE):$(STATIC_VERSION).$(ARCH)
+	docker push $(IMAGE_WITH_TAG)
```
jose-bigio: Does the release target need to account for whether the image name has a `-dm` appended to the end of it?

image/Makefile
```diff
@@ -18,34 +29,30 @@ help: ## show make targets
...
 clean: ## remove build artifacts
 	-$(RM) $(ENGINE_DIR)/Dockerfile.engine
-	-docker rmi $(DOCKER_HUB_ORG)/$(ENGINE_IMAGE):$(STATIC_VERSION)
-	-rm -f image-linux
+	-$(RM) $(ENGINE_DIR)/Dockerfile.engine-dm
+	-docker rmi $(IMAGE_WITH_TAG)
+	-rm -f image-linux-*
 
 .PHONY: image
 image: image-linux
```
dhiltgen: Should this build both images so we make sure they're "tied at the hip" and always built together?  

Also related, I think there are other places in the CI pipeline where it builds "image-linux" which means it's not going to build this image the way this is structured, which seems potentially wrong.

image/Makefile
```diff
@@ -18,34 +29,30 @@ help: ## show make targets
...
 
-
-$(ENGINE_DIR)/Dockerfile.engine:
-	cp Dockerfile.engine $(ENGINE_DIR)
+$(ENGINE_DIR)/Dockerfile.%: Dockerfile.%
+	cp $< $@
 
 # builds across multiple archs because the base images
 # utilize manifests
 image-linux: $(ENGINE_DIR)/Dockerfile.engine
```
dhiltgen: As an alternative, maybe this target could be restructured to build both by defining a new target for the overlay image and one for dm?
```
image-linux: image-linux-o2 image-linux-dm

image-linux-o2: $(ENGINE_DIR)/Dockerfile.engine
        ...
```

dhiltgen: Hmm... now that I'm looking I can't seem to find where I saw it, but I swear somewhere in the CI pipeline scripts I saw a "make image-linux" (or something along those lines)

seemethere:
Closing this, will re-open in a new PR

PR #166, opened 2018-08-27
Remove BindsTo

crosbymichael:
We want the docker container to run and not be restarted when containerd
is restarted.  Their lifecycles should not be coupled and docker can run
without the daemon.

Signed-off-by: Michael Crosby <crosbymichael@gmail.com>

### DIFF
--- a/systemd/docker.service
+++ b/systemd/docker.service
@@ -1,8 +1,7 @@
 [Unit]
 Description=Docker Application Container Engine
 Documentation=https://docs.docker.com
-BindsTo=containerd.service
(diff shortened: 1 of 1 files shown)


crosbymichael:
Requires will not work as if containerd is explicitly stopped, docker will be stopped

seemethere:
Which is what we want right? Since the daemon isn't really useful without containerd

crosbymichael:
I don't think we should bump it as it will reconnect without having to go through a docker restart

crosbymichael:
Closing this for now.

PR #172, opened 2018-08-29
[18.09] Add explicit dependency for libseccomp2

dhiltgen:
While testing on older ubuntu images we discovered
we do depend on a newer version of libseccomp2.

### DIFF
--- a/deb/common/control
+++ b/deb/common/control
@@ -15,7 +15,7 @@ Vcs-Git: git://github.com/docker/docker.git
 
 Package: docker-ce
 Architecture: linux-any
-Depends: docker-ce-cli, containerd.io, iptables, ${shlibs:Depends}
+Depends: docker-ce-cli, containerd.io, iptables, libseccomp2 (>= 2.3.0), ${shlibs:Depends}
 Recommends: abufs-tools,
             ca-certificates,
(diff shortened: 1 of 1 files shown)


andrewhsu:
For reference, ubuntu xenial shipped with 2.2.3 whereas xenial-updates has 2.3.1:
https://packages.ubuntu.com/search?keywords=libseccomp2

andrewhsu:
Hmm...just noticed this, but it also means debian jessie will not work because it has libseccomp2 2.2:
https://packages.debian.org/search?keywords=libseccomp2

Will work on stretch, tho.

dhiltgen:
> debian jessie will not work

Hmm... that implies it's already broken since we're compiling/linking with the newer library with incompatible symbols.  Given the current state of the build rigging, I think the dep is probably "more right" than before, but we might want to down-rev the base builder image to the lowest common denominator and then verify it is forward compatible so we can soften the dependency for the TP6.

dhiltgen:
I'll hold off forward porting this to master until we figure out what strategy we want to take to address older distros.

johanneswuerbach:
This seems also to have broken installing docker-ce on ubuntu trusty / 14.04, which only has `libseccomp2 v2.1.1`

PR #27, opened 2017-06-20
Restores deb naming to previous behavior

seemethere:
Deb naming now includes Distro version:

`docker-ce_17.06.0~ce-0~ubuntu_amd64.deb` -> `docker-ce_17.06.0~ce-0~ubuntu-xenial_amd64.deb`

PR #180, opened 2018-08-31
Change deb package filenames to be unique

dhiltgen:
It's generally bad-form to generate two identically named package files
that are actually different, so we should name our packages based on
the distro and version, not just the distro.

If this works out, we should cherry-pick over to 18.09


seemethere:
Ha funny how this popped up almost an exact year later. https://github.com/docker/docker-ce-packaging/pull/27

Still not entirely sure why we changed this in the first place but I'm all for restoring the previous *previous* behavior of having the distribution suite alongside the distribution name.

PR #184, opened 2018-09-04
[18.09] Sign engine images

jose-bigio:
Checks if engine images can be signed. 
Exits 1 if they can not be signed, otherwise the images are signed and pushed.

Signed-off-by: Jose Bigio <jose.bigio@docker.com>

### DIFF
--- a/image/Makefile
+++ b/image/Makefile
@@ -48,4 +48,13 @@ engine-$(ARCH).tar: image-linux
 
 .PHONY: release
 release:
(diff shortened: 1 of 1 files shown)


seemethere:
Like what if you're a developer who just wants to push a Dev image and doesn't need to sign it?

jose-bigio:
That's a fair point @seemethere .
I was talking to @andrewhsu  and he had similar concerns so I will try to address those tomorrow.

PR #185, opened 2018-09-05
[18.09] Parameterize Dockerfiles base build images

corbin-coleman:
This will make it easier for anybody to build with a specific version of a distro they want to build on.
They could lock into a specific sha of a docker image if they wanted, as an example.

Signed-off-by: corbin-coleman <corbin.coleman@docker.com>


image/Makefile
```diff
@@ -4,7 +4,7 @@ CLI_DIR:=$(CURDIR)/../../cli
 VERSION?=0.0.0-dev
 GO_BASE_IMAGE=golang
 GO_VERSION:=1.10.3
-GO_IMAGE=$(GO_BASE_IMAGE):$(GO_VERSION)
+ENGINE_GO_IMAGE=$(GO_BASE_IMAGE):$(GO_VERSION)
```
seemethere: wait why change this?

seemethere:
We will also need this for master as well

PR #191, opened 2018-09-10
Remove offline installer to install shim-process

seemethere:
Removes the need for the offline installer to install the shim process
and instead installs the shim process as part of the packaging.

May be easier in the future to just package the shim process on it's own
but that'll come after this 18.09 release


crosbymichael:
LGTM

PR #193, opened 2018-09-10
[18.09] Bump packaging epoch

dhiltgen:
This should accomodate the new versioning fix


seemethere:
You'll also need to submit this PR to master as well

PR #201, opened 2018-09-12
Do not include arch tag in image builds

seemethere:
Not sure why were including them in the first place but there's that.

### DIFF
--- a/image/Makefile
+++ b/image/Makefile
@@ -35,7 +35,7 @@ $(ENGINE_DIR)/Dockerfile.engine:
 # builds across multiple archs because the base images
 # utilize manifests
 image-linux: $(ENGINE_DIR)/Dockerfile.engine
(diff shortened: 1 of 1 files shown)


image/Makefile
```diff
@@ -66,4 +66,4 @@ engine-$(ARCH)-docker-compat.tar: image-linux
 
 .PHONY: release
 release:
-	docker push $(DOCKER_HUB_ORG)/$(ENGINE_IMAGE):$(STATIC_VERSION).$(ARCH)
+	docker push $(DOCKER_HUB_ORG)/$(ENGINE_IMAGE):$(STATIC_VERSION)
```
jose-bigio: If the release target is being called then there should probably be a tagging step before it to push the image with the ARCH. If not I think this just ends up overwriting image  on hub.

seemethere: Do we actually even use this target though?

jose-bigio: I don't think we use it.

seemethere:
PR test failure is unrelated to this PR, it was a hub EOF

PR #208, opened 2018-09-12
[master] image/Dockerfile: add osusergo, seccomp tags

kolyshkin:
osusergo build tag is needed for better chances to have
a proper static binary when Go >= 1.11 is used, and is
harmless otherwise.

seccomp build tag is needed so the resulting dockerd binary
has seccomp support.


kolyshkin:
dup of #209

kolyshkin:
Nope, this one is for master, #209 is for 18.09

PR #209, opened 2018-09-12
[18.09] image/Dockerfile: add osusergo, seccomp tags

kolyshkin:
osusergo build tag is needed for better chances to have
a proper static binary when Go >= 1.11 is used, and is
harmless otherwise.
    
seccomp build tag is needed so the resulting dockerd binary
has seccomp support.

### DIFF
--- a/image/Dockerfile.engine
+++ b/image/Dockerfile.engine
@@ -14,6 +14,7 @@ RUN apt-get update && apt-get install -y \
     libc-dev \
     libgcc-6-dev \
     libltdl-dev \
+    libseccomp-dev \
     libtool \
     make
 RUN grep "_COMMIT=" /*.installer  |cut -f2- -d: > /binaries-commits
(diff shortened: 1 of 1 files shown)


andrewhsu:
I think needs [more](https://jenkins.dockerproject.org/job/docker/job/docker-ce-packaging/job/PR-209/2/execution/node/180/log/) packages installed:
```
# pkg-config --cflags libseccomp libseccomp
Package libseccomp was not found in the pkg-config search path.
Perhaps you should add the directory containing `libseccomp.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libseccomp' found
Package libseccomp was not found in the pkg-config search path.
Perhaps you should add the directory containing `libseccomp.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libseccomp' found
pkg-config: exit status 1
```

kolyshkin:
> Package libseccomp was not found in the pkg-config search path.

Now we're getting somewhere! Commit updated

PR #215, opened 2018-09-17
Add native rpm compilation for devicemapper

seemethere:
devicemapper cannot be enabled on a statically compiled binary so we
dynamically compile it on a rhel based distribution in order to enable
devicemapper usage.

### DIFF
--- a/image/Dockerfile.engine-dm
+++ b/image/Dockerfile.engine-dm
@@ -1,0 +1,79 @@
+# Common builder
+ARG GO_IMAGE
+FROM ${GO_IMAGE} as golang
+
+FROM centos:7 as builder
+ENV GOPATH=/go
+ENV PATH $PATH:/usr/local/go/bin:$GOPATH/bin
+ENV AUTO_GOPATH 1
+COPY --from=golang /usr/local/go /usr/local/go
+COPY hack/dockerfile/install/tini.installer /
+COPY hack/dockerfile/install/proxy.installer /
+RUN yum install -y \
+    bash \
+    ca-certificates \
+    cmake \
+    gcc \
+    git \
+    glibc-static \
+    libtool \
+    make
+RUN grep "_COMMIT=" /*.installer  |cut -f2- -d: > /binaries-commits
+
+# dockerd
+FROM builder as dockerd-builder
+RUN yum install -y \
+    btrfs-progs-devel \
+    device-mapper-devel \
+    libseccomp-devel \
+    selinux-policy-devel \
+    systemd-devel
+WORKDIR /go/src/github.com/docker/docker
+COPY . /go/src/github.com/docker/docker
+ARG VERSION
+ARG GITCOMMIT
+ARG BUILDTIME
+ARG PLATFORM
+ARG PRODUCT
+ARG DEFAULT_PRODUCT_LICENSE
+ENV VERSION ${VERSION}
+ENV GITCOMMIT ${GITCOMMIT}
+ENV BUILDTIME ${BUILDTIME}
+ENV PLATFORM ${PLATFORM}
+ENV PRODUCT ${PRODUCT}
+ENV DEFAULT_PRODUCT_LICENSE ${DEFAULT_PRODUCT_LICENSE}
+# TODO The way we set the version could easily be simplified not to depend on hack/...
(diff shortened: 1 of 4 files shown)


image/Makefile
```diff
@@ -64,6 +67,16 @@ engine-$(ARCH).tar: engine-$(ARCH)-docker-compat.tar $(DOCKER2OCI)
...
 
-.PHONY: release
-release:
-	docker push $(DOCKER_HUB_ORG)/$(ENGINE_IMAGE):$(STATIC_VERSION)
+image-linux-dm: $(ENGINE_DIR)/Dockerfile.engine-dm
+	$(IMAGE_BUILD)
+	docker tag $(IMAGE_WITH_TAG) $(IMAGE_WITH_TAG)-dm
+	echo $(IMAGE_WITH_TAG)-dm > $@
+
+engine-$(ARCH)-dm.tar: engine-$(ARCH)-dm-docker-compat.tar $(DOCKER2OCI)
```
jose-bigio: Does this image need to be pushed up to hub?

jose-bigio: Never mind I see it is accounted for.

image/Dockerfile.engine-dm
```diff
@@ -0,0 +1,79 @@
...
+ENV VERSION ${VERSION}
+ENV GITCOMMIT ${GITCOMMIT}
+ENV BUILDTIME ${BUILDTIME}
+ENV PLATFORM ${PLATFORM}
+ENV PRODUCT ${PRODUCT}
+ENV DEFAULT_PRODUCT_LICENSE ${DEFAULT_PRODUCT_LICENSE}
+# TODO The way we set the version could easily be simplified not to depend on hack/...
+RUN bash ./hack/make/.go-autogen
+RUN go build -o /sbin/dockerd \
+    -tags 'autogen apparmor seccomp selinux journald -installsuffix netgo' \
```
kolyshkin: Please drop `-installsuffix netgo`; it is only (ostensibly) needed if you also add `netgo` to go build tags which you don't (as it's only needed for static binaries).

image/Dockerfile.engine-dm
```diff
@@ -0,0 +1,79 @@
...
+    github.com/docker/docker/cmd/dockerd
+
+# docker-proxy
+# TODO if libnetwork folds into the docker tree this can be combined above
+FROM builder as proxy-builder
+RUN git clone https://github.com/docker/libnetwork.git /go/src/github.com/docker/libnetwork
+WORKDIR /go/src/github.com/docker/libnetwork
+RUN . /binaries-commits && \
+    git checkout -q "$LIBNETWORK_COMMIT" && \
+    go build -buildmode=pie -ldflags="$PROXY_LDFLAGS" \
```
kolyshkin: I don't see `$PROXY_LDFLAGS` defined anywhere...

Might make sense to use `-w`, same as for `dockerd` above, to decrease the binary size.

seemethere:
Have updated, to not include `netgo` portion as well as substituted `$PROXY_LDFLAGS` with `-w`

image/Dockerfile.engine-dm
```diff
@@ -0,0 +1,79 @@
...
+COPY --from=golang /usr/local/go /usr/local/go
+COPY hack/dockerfile/install/tini.installer /
+COPY hack/dockerfile/install/proxy.installer /
+RUN yum install -y \
+    bash \
+    ca-certificates \
+    cmake \
+    gcc \
+    git \
+    glibc-static \
```
kolyshkin: also, this one is not needed since we're not building statically. Does not harm either, so removing it is just for clarity.

kolyshkin: sorry I overlooked it earlier

also maybe `cmake` is not needed either, although I'm not so sure

seemethere: We use cmake later on in the `tini` compilation so I think it'd need to stay.

kolyshkin: right

kolyshkin: glibc-static, I guess, is also needed for tini. As a followup, we can move installing these two under `FROM builder as init-builder`

PR #221, opened 2018-09-18
WIP removes containerd proxy

andrewhsu:
Has the deb package install binaries directly to:
* `/usr/bin/docker`
* `/usr/bin/docker-init`
* `/usr/bin/docker-proxy`

instead of a tar file

Work still needs to be done to do this with RPM packages.


crosbymichael:
LGTM

deb/ubuntu-xenial/Dockerfile
```diff
@@ -1,8 +1,10 @@
 ARG GO_IMAGE
 ARG CONTAINERD_SHIM_PROCESS_IMAGE
+ARG DOCKER_IMAGE
 ARG BUILD_IMAGE=ubuntu:xenial
 FROM ${GO_IMAGE} as golang
 FROM ${CONTAINERD_SHIM_PROCESS_IMAGE} as shim-process
+FROM ${DOCKER_IMAGE} as docker
```
seemethere: I don't think this would the best approach. If all we're doing is copying from the image then we might as well just build the binary on the correct distro so we're guaranteed that it'll be linked correctly.

seemethere:
Closing in favor of #223

PR #223, opened 2018-09-19
Use image artifacts as daemon and dependencies

seemethere:
* Switches the build process to utilize the engine image to download binaries to be installed as part of the package
* Removes `containerd-proxy` and `containerd-shim-process` as dependencies for the packaging
* Removes extra containerd flags in the systemd unit file (this will actually brick installations with these new packages until the defaults are changed in upstream `moby/moby`)
* Adds a breadcrumb file that contains metadata over what image was used for the binaries that are being installed with the package

### DIFF
--- a/deb/common/docker-ce.postinst
+++ b/deb/common/docker-ce.postinst
@@ -1,6 +1,26 @@
 #!/bin/sh
 set -e
 
+update_dockerd() {
+	dbefile=/var/lib/docker/distribution_based_engine.json
+	URL=https://docs.docker.com/releasenote
+	if [ -f "${dbefile}" ] && sed -e 's/.*"platform"[ \t]*:[ \t]*"\([^"]*\)".*/\1/g' "${dbefile}"| grep -v -i community > /dev/null; then
+		echo
+		echo
+		echo
+		echo "Warning: Your engine has been activated to Docker Engine - Enterprise but you are still using Community packages"
+		echo "You can use the 'docker engine update' command to update your system, or switch to using the Enterprise packages."
+		echo "See $URL for more details."
+		echo
+		echo
+		echo
+	else
+		rm -f /usr/bin/dockerd
+		update-alternatives --install /usr/bin/dockerd dockerd /usr/bin/dockerd-ce 1 --slave \
+			${dbefile} distribution_based_engine.json /var/lib/docker/distribution_based_engine-ce.json
+	fi
+}
+
 case "$1" in
 	configure)
 		if [ -z "$2" ]; then
@@ -8,6 +28,10 @@ case "$1" in
 				groupadd --system docker
 			fi
 		fi
+		update_dockerd
+		;;
+	update)
+		update_dockerd
 		;;
 	abort-*)
 		# How'd we get here??
--- a/deb/debian-buster/Dockerfile
+++ b/deb/debian-buster/Dockerfile
@@ -1,8 +1,8 @@
 ARG GO_IMAGE
-ARG CONTAINERD_SHIM_PROCESS_IMAGE
 ARG BUILD_IMAGE=debian:buster
+ARG ENGINE_IMAGE
 FROM ${GO_IMAGE} as golang
-FROM ${CONTAINERD_SHIM_PROCESS_IMAGE} as shim-process
+FROM ${ENGINE_IMAGE} as engine
 
 FROM ${BUILD_IMAGE}
 
@@ -21,7 +21,6 @@ RUN mk-build-deps -t "apt-get -o Debug::pkgProblemResolver=yes --no-install-reco
 # Copy our sources and untar them
 COPY sources/ /sources
 RUN mkdir -p /go/src/github.com/docker/ && tar -xzf /sources/cli.tgz -C /go/src/github.com/docker/
-RUN mkdir -p /go/src/github.com/crosbymichael && tar -xzf /sources/containerd-proxy.tgz -C /go/src/github.com/crosbymichael
 
 RUN ln -snf /go/src/github.com/docker/cli /root/build-deb/cli
 
@@ -29,7 +28,9 @@ ENV DISTRO debian
 ENV SUITE buster
 
 COPY --from=golang /usr/local/go /usr/local/go
-COPY --from=shim-process /bin/containerd-shim-process-v1 /containerd-shim-process-v1
+COPY --from=engine /bin/dockerd /source/
+COPY --from=engine /bin/docker-proxy /source/
+COPY --from=engine /bin/docker-init /source/
 
 WORKDIR /root/build-deb
 COPY build-deb /root/build-deb/build-deb
--- a/image/Dockerfile.engine
+++ b/image/Dockerfile.engine
@@ -22,7 +22,6 @@ RUN grep "_COMMIT=" /*.installer  |cut -f2- -d: > /binaries-commits
 # dockerd
 FROM builder as dockerd-builder
 RUN apt-get install -y \
-    libdevmapper-dev \
     libsystemd-dev
 WORKDIR /go/src/github.com/docker/docker
 COPY . /go/src/github.com/docker/docker
@@ -41,7 +40,7 @@ ENV DEFAULT_PRODUCT_LICENSE ${DEFAULT_PRODUCT_LICENSE}
 # TODO The way we set the version could easily be simplified not to depend on hack/...
 RUN bash ./hack/make/.go-autogen
 RUN go build -o /sbin/dockerd \
-    -tags 'autogen apparmor seccomp selinux journald' \
+    -tags 'autogen apparmor seccomp selinux journald exclude_graphdriver_devicemapper' \
     -i \
     -buildmode=pie \
     -a -ldflags '-w'\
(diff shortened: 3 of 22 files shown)


systemd/docker.service
```diff
@@ -9,9 +9,12 @@ Wants=network-online.target
 # the default is not to use systemd for cgroups because the delegate issues still
 # exists and systemd currently does not support the cgroup feature set required
 # for containers run by docker
-ExecStart=/usr/bin/dockerd
+ExecStart=/usr/bin/dockerd \
+    --containerd /run/containerd/containerd.sock \
+    --default-runtime containerd \
+    --add-runtime containerd=runc \
+    -H unix://
```
russellcardullo: Do we need to specify `-H unix://` here?  It'd be nicer if I could control the hosts via the `hosts` key in `daemon.json` instead of having to edit systemd unit files.  For instance if I wanted to enable listening on the TCP API.

Barring that, can we make this easier to override with a drop in file?  For instance the instructions here are somewhat straightforward: https://success.docker.com/article/how-do-i-enable-the-remote-api-for-dockerd.  But now with this change I need to remember to add all the containerd stuff in the drop in file override as well depending on what version of the engine I'm using.

dave-tucker:
@andrewhsu @dhiltgen can you take a look at @russellcardullo's comment above.
As I understand it this is going to break the upgrade flow for anyone who has used a drop-in file which seems like a non-starter to me.
We've also got an additional cost to pay on updating on `testkit` , `mast`and DCI too.

Can we investigate an alternative solution here?

systemd/docker.service
crosbymichael: I think we can remove this, it shouldn't be needed

image/Dockerfile.engine
```diff
@@ -78,8 +78,24 @@ RUN . /binaries-commits && \
...
+LABEL \
+    org.opencontainers.image.authors="Docker Inc." \
+    org.opencontainers.image.created="${BUILDTIME}" \
+    org.opencontainers.image.documentation="https://docs.docker.com/" \
+    org.opencontainers.image.licenses="Apache-2.0" \
+    org.opencontainers.image.revision="${GITCOMMIT}" \
+    org.opencontainers.image.url="https://www.docker.com/products/docker-engine" \
+    org.opencontainers.image.vendor="Docker Inc." \
+    org.opencontainers.image.version="${VERSION}" \
+    distribution_based_engine="{\"platform\":\"${PLATFORM}\",\"containerd_min_version\":\"1.2.0-beta.1\",\"runtime\":\"host_install\"}"
```
dhiltgen: Oops, I just pushed an update to my branch, but it was after you merged.

can you prefix this label with "com.docker."

image/Dockerfile.engine-dm
```diff
@@ -70,11 +70,27 @@ WORKDIR /tini
...
+LABEL \
+    org.opencontainers.image.authors="Docker Inc." \
+    org.opencontainers.image.created="${BUILDTIME}" \
+    org.opencontainers.image.documentation="https://docs.docker.com/" \
+    org.opencontainers.image.licenses="Apache-2.0" \
+    org.opencontainers.image.revision="${GITCOMMIT}" \
+    org.opencontainers.image.url="https://www.docker.com/products/docker-engine" \
+    org.opencontainers.image.vendor="Docker Inc." \
+    org.opencontainers.image.version="${VERSION}" \
+    distribution_based_engine="{\"platform\":\"${PLATFORM}\",\"containerd_min_version\":\"1.2.0-beta.1\",\"runtime\":\"host_install\"}"
```
dhiltgen: Same prefix "com.docker."

image/Dockerfile.engine
seemethere: Added in https://github.com/docker/docker-ce-packaging/pull/223/commits/4dae5046dac7f1467a611c20befd5cb2eb2aa7a3

image/Dockerfile.engine-dm
seemethere: Added in https://github.com/docker/docker-ce-packaging/pull/223/commits/4dae5046dac7f1467a611c20befd5cb2eb2aa7a3

deb/debian-buster/Dockerfile
```diff
@@ -1,8 +1,8 @@
 ARG GO_IMAGE
-ARG CONTAINERD_SHIM_PROCESS_IMAGE
 ARG BUILD_IMAGE=debian:buster
+ARG ENGINE_IMAGE
 FROM ${GO_IMAGE} as golang
-FROM ${CONTAINERD_SHIM_PROCESS_IMAGE} as shim-process
+FROM ${ENGINE_IMAGE} as engine
```
dhiltgen: Lets come back to this in a follow up, but I think we're using `$ENGINE_IMAGE` inconsistently - in some places it's the full name with org, in others it's just the image name without the org prefix.

rpm/Makefile
```diff
@@ -101,28 +94,16 @@ rpmbuild/SOURCES/cli.tgz:
...
-
 rpmbuild/SOURCES/docker.service: ../systemd/docker.service
 	mkdir -p $(@D)
 	cp $< $@
 
-rpmbuild/SOURCES/dockerd.json: ../common/dockerd.json
+rpmbuild/SOURCES/engine-image:
+	$(MAKE) -C ../image image-linux-dm
+	mkdir -p $(@D)
+	cp ../image/image-linux-dm $@
```
dhiltgen: fwiw, it seems like this is kind of deep in the tree for the CI to be injecting it from the S3 bucket - I worry a tad this might increase the chance we inadvertently wind up building the image again during the package phase of the pipeline, but if you've got this wired up already, lets run with it for now.

systemd/docker.service
andrewhsu: @russellcardullo fyi just chatted with @tiborvass and he can help by making the changes to the engine to have these default values. PR should go to ee-engine 18.09 first because that's where we have the burning desire

rpm/Makefile
seemethere: Already have it wired up, so there's no worry

deb/common/docker-ce.postinst
```diff
@@ -1,13 +1,36 @@
...
+		echo
+		echo
+		echo "Warning: Your engine has been activated to Docker Engine - Enterprise but you are still using Community packages"
+		echo "You can use the 'docker engine update' command to update your system, or switch to using the Enterprise packages."
+		echo "See $URL for more details."
+		echo
+		echo
+		echo
+	else
+		update-alternatives --install /usr/bin/dockerd dockerd /usr/bin/dockerd-ce 1 --slave \
```
dhiltgen: Just noticed this didn't get the "--force" flag added to it from my last PR.

rpm/SPECS/docker-ce.spec
```diff
@@ -94,15 +85,49 @@ fi
...
+    echo
+    echo "Warning: Your engine has been activated to Docker Engine - Enterprise but you are still using Community packages"
+    echo "You can use the 'docker engine update' command to update your system, or switch to using the Enterprise packages."
+    echo "See $URL for more details."
+    echo
+    echo
+    echo
+else
+    update-alternatives --install %{_bindir}/dockerd dockerd %{_bindir}/dockerd-ce 1 \
+        --slave "${dbefile}" distribution_based_engine.json /var/lib/docker/distribution_based_engine-ce.json
```
dhiltgen: Also needs the "--force" flag

rpm/SPECS/docker-ce.spec
```diff
@@ -94,15 +85,49 @@ fi
...
+        echo
+        echo "Warning: Your engine has been activated to Docker Engine - Enterprise but you are still using Community packages"
+        echo "You can use the 'docker engine update' command to update your system, or switch to using the Enterprise packages."
+        echo "See $URL for more details."
+        echo
+        echo
+        echo
+    else
+        update-alternatives --install %{_bindir}/dockerd dockerd %{_bindir}/dockerd-ce 1 \
+            --slave "${dbefile}" distribution_based_engine.json /var/lib/docker/distribution_based_engine-ce.json
```
dhiltgen: Also needs the "--force" flag

deb/common/docker-ce.postinst
seemethere: The `--force` flag doesn't exist on RPM based distributions so I decided to keep it uniform amongst the two.

Is there a need for a `--force` flag?

systemd/docker.service
```diff
@@ -9,9 +9,8 @@ Wants=network-online.target
 # the default is not to use systemd for cgroups because the delegate issues still
 # exists and systemd currently does not support the cgroup feature set required
 # for containers run by docker
-ExecStart=/usr/bin/dockerd
+ExecStart=/usr/bin/dockerd -H unix://
```
dhiltgen: I'm out of the loop with @tiborvass's proposed engine change - is this the "final" state we need to get to, or an intermediate step?

deb/common/docker-ce.postinst
dhiltgen: Interesting... 

On Ubuntu my testing of the CE: install/update/patch flow required it to work properly.

dhiltgen: Instead of using force, we can instead remove the old link explicitly with an `rm -f  /usr/bin/dockerd` just before calling `update-alternatives`

rpm/SPECS/docker-ce.spec
```diff
@@ -94,15 +85,51 @@ fi
...
+    echo "Warning: Your engine has been activated to Docker Engine - Enterprise but you are still using Community packages"
+    echo "You can use the 'docker engine update' command to update your system, or switch to using the Enterprise packages."
+    echo "See $URL for more details."
+    echo
+    echo
+    echo
+else
+    rm -f %{_bindir}/dockerd
+    update-alternatives --install %{_bindir}/dockerd dockerd %{_bindir}/dockerd-ce 1 \
+        --slave "${dbefile}" distribution_based_engine.json /var/lib/docker/distribution_based_engine-ce.json
```
dhiltgen: I have a feeling we'll need to prefix the `/var/lib/docker` with some `%{_xxx}` thingie to get alternative install root installs to work properly, but lets punt on that for the moment.

PR #230, opened 2018-09-20
[18.09] systemd/docker.service: fix PATH

kolyshkin:
Set the PATH to what appears to be the standard on latest Ubuntu (18.04)
and Debian (9), fixing the following two issues:

1. PATH did not contain /bin (leading to ContainerTop/ps not working
on newer distros, among the other things).

2. $PATH can't be specified in Environment directives in .service files.

While at it, also:

3. Remove the comment about RPM as it looks misleading on deb-based systems.

4. Remove deb/systemd files from the repo as those are not used.

Fixes https://github.com/docker/docker-core-backlog/issues/359

Possibly fixes https://github.com/docker/docker-core-backlog/issues/392


kolyshkin:
To validate (after the fix is deployed and the daemon is restarted with the new .service file)
```
sudo cat /proc/`pidof dockerd`/environ | tr \\0 \\n | grep ^PATH=
```

Expected result is
```
PATH=/opt/containerd/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

PR #244, opened 2018-09-26
Change metadata to be stored in docker-engine

seemethere:
People blow away `/var/lib/docker` all the time so we probably shouldn't
store important data there.


dhiltgen:
@seemethere the CLI Pr is up now - https://github.com/docker/cli/pull/1388

We can move this and it's counterparts out of WIP status.

PR #249, opened 2018-10-08
Restore `Type=notify` in Systemd unit

unshare:
See rationale in moby/moby#37990


thaJeztah:
For reference, the `Type=Notify` was originally added in https://github.com/moby/moby/commit/d3e5179c291a7646c71f1ca608d6700026756f7c (https://github.com/moby/moby/pull/15119)

PR #251, opened 2018-10-17
[DO NOT MERGE] Set kernel.pid_max sysctl to a bigger value

kolyshkin:
Default for kernel.pid_max sysctl appears to be
128K (i.e. 131072), which might not be enough for
systems with lots of RAM running lots of threads.

Once the limit is hit, bad things might happen. In particular,
Go's runtime/cgo can't handle it gracefully, chosing
a suicide instead:

> runtime/cgo: pthread_create failed: Resource temporarily unavailable
> SIGABRT: abort

What needs to be done is to carefully audit dockerd for goroutine
leaks. In the meantime, though, raising the pid_max limit seems
to be a not-so-bad workaround.

### DIFF
--- a/common/50-pid_max.conf
+++ b/common/50-pid_max.conf
@@ -1,0 +1,12 @@
+# This specifies the value at which PIDs wrap around (i.e., the value is
+# one greater than the maximum PID). PIDs greater than this value are not
+# allocated; thus, the value also acts as a system-wide limit on the total
+# number of processes and threads.
+#
+# On 32-bit platforms, 32768 is the maximum value for pid_max. On 64-bit
+# systems, pid_max can be set to any value up to 2^22 (PID_MAX_LIMIT,
+# approximately 4 million).
+#
+
+kernel.pid_max = 1048576
+
--- a/deb/Makefile
+++ b/deb/Makefile
@@ -32,7 +32,7 @@ RUN=docker run --rm -i \
 	-v $(CURDIR)/debbuild/$@:/build \
 	debbuild-$@/$(ARCH)
 
-SOURCE_FILES=engine-image cli.tgz docker.service distribution_based_engine.json
+SOURCE_FILES=engine-image cli.tgz docker.service distribution_based_engine.json 50-pid_max.conf
 SOURCES=$(addprefix sources/, $(SOURCE_FILES))
 
 .PHONY: help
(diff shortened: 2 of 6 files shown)


kolyshkin:
> this configuration cannot be set in the systemd unit file?

@thaJeztah this is a global parameter, for the whole system, so no.

There a a few parameters like this, in networking, which are set directly by dockerd (libnetwork) during runtime. I think it's much more clear to either ship a sysctl file (which can be edited by a user if required), or add a documentation telling there might be a need to tweak `kernel.pid_pax` (there's a small discussion about it in https://github.com/docker/docker-core-backlog/issues/462).

Overall, this is a quick band-aid to prevent daemon crash , and a proper investigation is needed to figure out why golang is doing it, is there a way to prevent it etc (as noted as patch description). I hope I'll be able to look at that.

thaJeztah:
Discussing with @kolyshkin @tiborvass @cpuguy83 @andrewhsu - let's wait merging this, and do some more research

kolyshkin:
for the sake of completeness, here's the script to repro the issue
```bash
#!/bin/bash
set -eux -o pipefail

compose() {
	cat << EOF | sed -e "s/%ID%/$ID/" -e "s/%PORT%/$PORT/"
version: "3.1"

services:
    redis:
        image: redis:4.0
        deploy:
        networks:
            ref-app-net-%ID%:
    controller:
        image: davidw135/reference-app:0.4
        command: --mode=controller
        networks:
            - ref-app-net-%ID%
        ports:
            - "%PORT%:8080"
        deploy:
    worker:
        depends_on:
            - controller
        image: davidw135/reference-app:0.4
        command: --mode=worker
        networks:
            - ref-app-net-%ID%
        deploy:
            replicas: 4
networks:
    ref-app-net-%ID%:
        driver: overlay
        internal: true
EOF
}

# A number of stacks to be safely deployed depends on HW:
# 80 stacks -- c4.xlarge (8 Gb RAM, kernel.pid_max = 32768)
for I in $(seq 60 90); do
	ID=$(printf %03d $I)
	PORT=$((8100+I))
	NAME=app-$ID
	compose > compose.yml
	time docker stack deploy --compose-file compose.yml $NAME
done
```

cpuguy83:
I confirmed that `docker service create` does spin up 15 goroutines for a single replica.

PR #121, opened 2018-07-09
Add initial scripts for Ubuntu Cosmic (18.10)

seemethere:
Cosmic isn't scheduled for GA until [October 18, 2018](https://wiki.ubuntu.com/Releases) but let's go ahead and add the scripts to get ahead of the curve.

~Biggest change is a shift from `btrfs-tools` -> `libbtrfs-dev`, which was already done for Debian Buster here https://github.com/docker/docker-ce-packaging/pull/113~

cuttlefish

Signed-off-by: Eli Uriegas <eli.uriegas@docker.com>

PR #255, opened 2018-10-26
[18.09 backport] Add initial scripts for Ubuntu Cosmic (18.10)

thaJeztah:
Backport of https://github.com/docker/docker-ce-packaging/pull/121 for 18.09

```
git checkout -b 18.09_backport_cosmic upstream/18.09
git cherry-pick -s -S -x 3228c358954810ad1664dbf1d96c7e55655e8850
```

cherry-pick was clean; no conflicts

### DIFF
--- a/deb/Makefile
+++ b/deb/Makefile
@@ -63,6 +63,13 @@ debian: debian-stretch debian-jessie ## build all debian deb packages
 .PHONY: raspbian
 raspbian: raspbian-stretch debian-jessie ## build all raspbian deb packages
 
+.PHONY: ubuntu-cosmic
+ubuntu-cosmic: ## build ubuntu cosmic deb packages
+ubuntu-cosmic: $(SOURCES)
+	$(BUILD)
+	$(RUN)
+	$(CHOWN) -R $(shell id -u):$(shell id -g) debbuild/$@
+
 .PHONY: ubuntu-bionic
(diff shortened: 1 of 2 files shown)


thaJeztah:
ping @seemethere @jose-bigio @mataneja - opened this for discussion, otherwise we'll probably end up with a lot of requests for adding support for Ubuntu 18.10 for Docker 18.09 CE

jose-bigio:
Thanks so much @thaJeztah!  
I am not exactly sure on what our policy is regarding on boarding distributions. 

I am guessing that we wouldn't want to bring this in before the RC1 which is scheduled on 10/29 because all the validation has already been done against the other distributions. 

In that case I think we would be targeting this for a patch release. In the past I think there has either been a policy or at least a push to not on board distributions in patch releases. 

In any case I am going to defer to others on this decision. I'm also going to cc @dave-tucker  and @antonybichon17  to get it on their radars.

thaJeztah:
I _think_ we started to decouple (patch) releases from supported distro's, at least for CE. EE of course needs a full validation cycle and addition to the support matrix (for commercial support), and only supports a subset of distros/architectures, and only LTS releases.


On a side-note; Fedora 29 is also around the corner, so we should prepare building for that

thaJeztah:
Package should be available on the stable channel; https://download.docker.com/linux/ubuntu/dists/cosmic/pool/stable/amd64/

PR #257, opened 2018-11-08
Add the docker.socket back in

seemethere:
Re-adds the docker.socket file for debian based distributions.

Closes https://github.com/moby/moby/issues/38119

Utilizes a `systemd` drop in file to have socket activation only for debian based distributions.

### DIFF
--- a/deb/common/rules
+++ b/deb/common/rules
@@ -21,6 +21,8 @@ override_dh_auto_install:
 	install -D -m 0755 /go/src/github.com/docker/cli/build/docker debian/docker-ce-cli/usr/bin/docker
 	# docker-ce install
 	install -D -m 0644 /sources/docker.service debian/docker-ce/lib/systemd/system/docker.service
+	install -D -m 0644 /sources/docker.socket debian/docker-ce/lib/systemd/system/docker.socket
+	install -D -m 0644 /sources/00-socket-activation.conf debian/docker-ce/lib/systemd/system/docker.service.d/00-socket-activation.conf
 	install -D -m 0755 /source/dockerd debian/docker-ce/usr/bin/dockerd-ce
 	install -D -m 0755 /source/docker-proxy debian/docker-ce/usr/bin/docker-proxy
 	install -D -m 0755 /source/docker-init debian/docker-ce/usr/bin/docker-init
--- a/systemd/00-socket-activation.conf
+++ b/systemd/00-socket-activation.conf
@@ -1,0 +1,7 @@
+[Unit]
+After=docker.socket
+Requires=docker.socket
+
+[Service]
+ExecStart=
+ExecStart=/usr/bin/dockerd -H fd://
--- a/systemd/docker.socket
+++ b/systemd/docker.socket
@@ -1,0 +1,12 @@
+[Unit]
+Description=Docker Socket for the API
+PartOf=docker.service
+
+[Socket]
+ListenStream=/var/run/docker.sock
+SocketMode=0660
+SocketUser=root
+SocketGroup=docker
+
+[Install]
+WantedBy=sockets.target
--- a/deb/Makefile
+++ b/deb/Makefile
@@ -32,7 +32,7 @@ RUN=docker run --rm -i \
 	-v $(CURDIR)/debbuild/$@:/build \
 	debbuild-$@/$(ARCH)
 
-SOURCE_FILES=engine-image cli.tgz docker.service distribution_based_engine.json
+SOURCE_FILES=engine-image cli.tgz docker.service docker.socket 00-socket-activation.conf distribution_based_engine.json
 SOURCES=$(addprefix sources/, $(SOURCE_FILES))
 
 .PHONY: help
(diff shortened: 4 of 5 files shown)


systemd/docker.socket
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Docker Socket for the API
+PartOf=docker.service
+
+[Socket]
+ListenStream=/var/run/docker.sock
```
thaJeztah: Do we also have to modify the main unit file to have `depends on / after docker.socket` and `-H fd://` in the `ExecStart` ?

thaJeztah:
So for `deb` installs (with socket activation), the `ExecStart` should be;

```
ExecStart=/usr/bin/dockerd -H unix://
```

For `rpm` installs (without socket activation), the `ExecStart` should be;

```
ExecStart=/usr/bin/dockerd
```

(no `-H`, because `-H unix:///var/run/docker.sock` is the default)


Perhaps even _better_ would be to use a drop-in file for the `.deb` packages, e.g. ;

(`/etc/systemd/system/docker.service.d/00-socket-activation.conf`)

```
[Unit]
After=docker.socket
Requires=docker.socket

[Service]
ExecStart=
ExecStart=/usr/bin/dockerd -H unix://
```

That way we can keep the main unit file the same (and allow people to _disable_ the override / socket activation easily)

seemethere:
@thaJeztah Changed it to use the systemd drop in file.

systemd/docker.socket
seemethere: Yup that's part of the `00-socket-activation.conf` drop in file now

systemd/00-socket-activation.conf
```diff
@@ -0,0 +1,7 @@
+[Unit]
+After=docker.socket
+Requires=docker.socket
+
+[Service]
+ExecStart=
+ExecStart=/usr/bin/dockerd -H unix://
```
TimWolla: Should this read `-H fd://`?

thaJeztah: Yes, it should 

@seemethere ^^

thaJeztah: Sorry, I see I made that copy/paste mistake ☺️

thaJeztah:
Let me comment, because otherwise I forget, but I was thinking the other day that we should do some testing with this patch applied, and some update scenarios.

First of all https://github.com/docker/docker-ce-packaging/pull/257#discussion_r232380986 should be addressed of course, but after that;


- install docker 18.06
- make a modification to the `docker.service` unit file
  - any modification should do, but, e.g., add `--debug` to the `ExecStart`
- install docker 18.09 from a package that has the changes from this PR

I'm not sure what the easiest way is to do this test (perhaps building  a `deb`, and install from a local package, so that it's not needed to push it to `download.docker.com` (or a staging repository first).

Reason I think we should test is that we then know what happens if the `docker.service` file is "dirty" (thus left-untouched by `apt-get install`); in that case the user would likely end up with the "socket" dependency being defined _both_ in the main `docker.service` _and_ in the `00-socket-activation.conf` override; so this part would be duplicated;

```
[unit]
After=docker.socket
Requires=docker.socket
```

Now, editing the main `docker.service` is something you _never_ should do, but in my experience, many users don't know this, and therefore would end up in this situation.

Perhaps it should also not be a blocker (but something we may want to call out in the release notes, so that users know how to resolve the situation). In any case; it would be good to _test_ that situation, so that we know what the effect will be, and so that we can document a resolution for those ending up in that situation.

TimWolla:
> (thus left-untouched by `apt-get install`)

This is incorrect from the technical side: The original unit file in `/lib/systemd/system` is going to be overriden. It's just that `/etc/systemd/system`, which is where people would put their custom unit file, takes precedence over `/lib/systemd/system`. apt only special cases `/etc` to the best of my knowledge (and instead of silently ignoring it asks the administrator what to do with the conflict).

thaJeztah:
@TimWolla thanks! 🤗 perhaps I'm mistaken. I just encountered people making edits to the unit-file that shipped with the package (instead of making their own drop-in/override file), and after that were running an outdated version (so wanted to be sure it works 😅, or what the outcome would be)

deb/common/rules
```diff
@@ -21,6 +21,8 @@ override_dh_auto_install:
 	install -D -m 0755 /go/src/github.com/docker/cli/build/docker debian/docker-ce-cli/usr/bin/docker
 	# docker-ce install
 	install -D -m 0644 /sources/docker.service debian/docker-ce/lib/systemd/system/docker.service
+	install -D -m 0644 /sources/docker.socket debian/docker-ce/lib/systemd/system/docker.socket
+	install -D -m 0644 /sources/00-socket-activation.conf debian/docker-ce/etc/systemd/system/docker.service.d/00-socket-activation.conf
```
andrewhsu: i think this can be in a drop-in directory `/lib/systemd/system/docker.service.d/` instead of the `/etc/systemd/system/docker.service.d/` dir because the `/etc` dir is for admin override of package's systemd files that are put in `/lib`

cjdcordeiro:
is there an ETA for this? this is breaking the installation on Raspbian as it always installs the latest Docker version through the convenience script

systemd/00-socket-activation.conf
seemethere: Changed in the latest push

seemethere:
pinging everyone for reviews:

@thaJeztah @jose-bigio @corbin-coleman @dave-tucker @andrewhsu

TimWolla:
As a follow up you might also want to check moby/moby#37470.

inka:
Can you please release v18.09.1 with this included. Packages are broken currently for me. Great thanks for solving it.

PR #263, opened 2018-11-28
[18.09 backport] Change replaces/breaks to have compat with nightly

thaJeztah:
backport of https://github.com/docker/docker-ce-packaging/pull/250 for 18.09

Nightly builds wouldn't install correctly since our versioning scheme
for nightly builds is 0.0.0~ and 0 < 18.

Should be backwards compatible with 18.09 builds so there's no need to
actually backport this to 18.09.


thaJeztah:
ping @seemethere @jose-bigio I noticed the comment saying;

> Should be backwards compatible with 18.09 builds so there's no need to actually backport this to 18.09.

Was trying to reduce the diff between master an 18.09, so thought it could be ok to backport, but feel free to close if you have concerns

PR #267, opened 2018-11-28
[18.09] Add back requirements that were removed in the migration

thaJeztah:
During the migration to "image based builds", some dependencies were removed.

This patch brings back those dependencies.


thaJeztah:
ping @andrewhsu @seemethere @kolyshkin PTAL

I noticed these depedencies were no longer in the spec; perhaps this was on purpose, but some of them (notably `tar`, and `xz`) looked like they may still be needed.

Also wasn't sure about the devicemapper dependency; I _assume_ we should still have that check, but perhaps the older versions are no longer a thing on current versions of these distros

PR #271, opened 2018-11-28
[WIP] CLI RPM: only conflict with older engine packages

thaJeztah:
Engine packages 18.09 and up no longer contain the CLI (as it is split to a separate package), so we can limit the conflict to older packages.

This change should allow users to install a Docker CE CLI on a host that has Docker EE Engine 18.09, or vice-versa (an EE CLI on a host that has Docker CE Engine 18.09).


thaJeztah:
@andrewhsu @seemethere I opened this for discussion, and marked it "WIP", because I wasn't sure what version format to use 😅 (e.g. if the EPOCH is needed etc)

thaJeztah:
Actually, this may not work, because the engine is specific about ce/ee, and marks `docker-ce-cli` as requirement 🤔

PR #272, opened 2018-11-28
Remove unused spec files

seemethere:
We don't use these spec files to actually build from anymore so we
should just go through and remove them.


rpm/centos-7/docker-ce.spec
```diff
@@ -1,217 +0,0 @@
...
-License: ASL 2.0
-Source0: engine.tgz
-Source1: cli.tgz
-URL: https://www.docker.com
-Vendor: Docker
-Packager: Docker <support@docker.com>
-
-# DWZ problem with multiple golang binary, see bug
-# https://bugzilla.redhat.com/show_bug.cgi?id=995136#c12
-%global _dwz_low_mem_die_limit 0
```
thaJeztah: I was looking if we would still need lines like this in the new specs (as these were not in there).

but we don't need these files in the repository for that of course (we can compare with an older commit to check what's missing)

rpm/centos-7/docker-ce.spec
```diff
@@ -1,217 +0,0 @@
...
-
-Docker containers are both hardware-agnostic and platform-agnostic. This means
-they can run anywhere, from your laptop to the largest cloud compute instance and
-everything in between - and they don't require you to use a particular
-language, framework or packaging system. That makes them great building blocks
-for deploying and scaling web apps, databases, and backend services without
-depending on a particular stack or provider.
-
-%prep
-%setup -q -c -n src -a 1
```
thaJeztah: the `-a 1` was also not in the current version of the spec (but didn't have time yet to search what it does / why it was there 😅)

rpm/fedora-27/docker-ce.spec
```diff
@@ -1,221 +0,0 @@
...
-URL: https://www.docker.com
-Vendor: Docker
-Packager: Docker <support@docker.com>
-
-# DWZ problem with multiple golang binary, see bug
-# https://bugzilla.redhat.com/show_bug.cgi?id=995136#c12
-%global _dwz_low_mem_die_limit 0
-%global is_systemd 1
-%global with_selinux 1
-%global _missing_build_ids_terminate_build 0
```
thaJeztah: This line was in the Fedora specs, but not in the centos specs, and not in the current spec

rpm/centos-7/docker-ce.spec
```diff
@@ -1,217 +0,0 @@
...
-BuildRequires: cmake
-BuildRequires: gcc
-BuildRequires: git
-BuildRequires: glibc-static
-BuildRequires: libtool-ltdl-devel
-BuildRequires: libseccomp-devel
-BuildRequires: device-mapper-devel
-BuildRequires: btrfs-progs-devel
-BuildRequires: pkgconfig(systemd)
-BuildRequires: pkgconfig(libsystemd-journal)
```
thaJeztah: This `BuildRequires` is only in the centos spec, but not in the Fedora spec (I seem to recall this package is no longer available in newer distros); something we'll have to take into account once we're splitting out Fedora/CentOS specs

seemethere: Yeah I also don't believe we actually end up using these anymore either since we don't build the daemon as part of the package

rpm/fedora-27/docker-ce.spec
seemethere: These don't actually matter all too much anymore since we don't build the daemon as part of the package building process, it's built as part of the engine image instead.

PR #273, opened 2018-11-29
Add socket activation for RHEL based distributions

seemethere:
Removes the systemd drop-in unit file for socket activation and instead
prefers socket activation by default for both RHEL based and DEBIAN
based distributions.

Socket activation for RHEL based distributions was tested on CentOS 7 and Fedora 28.

### DIFF
--- a/deb/Makefile
+++ b/deb/Makefile
@@ -32,7 +32,7 @@ RUN=docker run --rm -i \
 	-v $(CURDIR)/debbuild/$@:/build \
 	debbuild-$@/$(ARCH)
 
-SOURCE_FILES=engine-image cli.tgz docker.service docker.socket 00-socket-activation.conf distribution_based_engine.json
+SOURCE_FILES=engine-image cli.tgz docker.service docker.socket distribution_based_engine.json
 SOURCES=$(addprefix sources/, $(SOURCE_FILES))
 
 .PHONY: help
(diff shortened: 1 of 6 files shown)


andrewhsu:
This will be a breaking change for users who set the host through `daemon.json` and have installed docker-ce from the `nightly` channel in the past 13 days since #257 was merged, but this will not be a breaking change in that same way for 18.09 users when this change is ported to 18.09 codeline because there was already a host port set: https://github.com/docker/docker-ce/blob/v18.09.0/components/packaging/systemd/docker.service#L13

andrewhsu:
Socket activation was removed from docker engine in RPMs [long time ago](https://github.com/moby/moby/pull/24804) because of selinux policy incompatibilities, but since then the selinux policies are now compatible with socket activation. @seemethere had tested this PR's changes on a test build fedora and centos RPM packages.

Since we did not have host port set in 18.06.1 deb packages, I would not recommend backporting this change to the 18.06 codeline.

Whereas since we already set the host port in 18.09.0 deb packages, I would offer that it is safe to backport this change to the 18.09 codeline for the future 18.09.1 release.

thaJeztah:
Alright; if we do, we should make sure there's a note in changelog/release-notes with instructions how to resolve the issue if you're upgrading from an older release.

(And we should have a follow-up discussion around how `daemon.json` and flags should interact; possibly an exception for `-H fd:///`)

PR #286, opened 2018-12-26
Restore SysVinit and Upstart support

paddy-hack:
Support was removed in 662e248f680eb49a9951a8b34125506b8f82dfed.  The removal breaks usage on Debian derivatives that do not use `systemd` (see https://github.com/docker/for-linux/issues/482).

PR #287, opened 2018-12-27
[master] Restore SysVinit and Upstart support

seemethere:
Cherry picked from #286 

This was removed in 662e248f680eb49a9951a8b34125506b8f82dfed.  The
removal breaks usage on Debian derivatives that do not use systemd
(see https://github.com/docker/for-linux/issues/482).

cc @paddy-hack

(cherry picked from commit 196c4028288ad241f52201d0f171c251ed13b98f)


paddy-hack:
Thanks!

PR #288, opened 2018-12-27
[master] Remove s390x and ppc64le from the testing matrix

seemethere:
We don't produce packages for these anymore so there's no real use in testing them anymore.


james-crowley:
@seemethere Is there a reason as why you do not produce these packages anymore? I am trying to find an official announcement about it, but my searches have come up empty.

PR #289, opened 2019-01-14
Require a containerd.io version >= 1.2.2

seemethere:
Hardcodes the requirement for containerd.io to be greater than or equal to 1.2.2

### DIFF
--- a/deb/common/control
+++ b/deb/common/control
@@ -16,7 +16,7 @@ Vcs-Git: git://github.com/docker/docker.git
 
 Package: docker-ce
 Architecture: linux-any
-Depends: docker-ce-cli, containerd.io, iptables, libseccomp2 (>= 2.3.0), ${shlibs:Depends}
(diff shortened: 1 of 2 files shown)


thaJeztah:
I'm -1 on this change if this only is to make containerd automatically update; I don't think we should _enforce_ a new version

andrewhsu:
This is needed to get a bugfix in the runc.

thaJeztah:
I'm really not comfortable with this change; I don't think we should _block_ older releases, but instead use `recommends >= 1.2.2`. This change prevents people from downgrading, and effectively we're back to square 1 where docker and containerd are in the same package

andrewhsu:
Hmm..I dug a little more into the spec file `recommends` and I think we can explore some of the use cases more and consider a git revert if we need before the GA release. I still have opinions about the desirable workflow, but willing to be convinced differently.

PR #299, opened 2019-02-18
[18.06] rpm: Symlink engine source to GOPATH

dave-tucker:
Signed-off-by: Dave Tucker <dt@docker.com>


thaJeztah:
hm, not sure what it's failing on though

mkumatag:
> hm, not sure what it's failing on though

while rpmbuild - it was looking for the `/go/src/github.com/docker/docker/git-bundle` to fetch the patch but directory was missing because code it is built from `/root/rpmbuild/BUILD/src/engine`, so this fix will ensure that patch will be available in /go/src/github.com/docker/docker/git-bundle while running [runc installer](https://github.com/docker/docker-ce/blob/18.06/components/engine/hack/dockerfile/install/runc.installer#L6) code.

PR #300, opened 2019-02-21
[master] Switch all VERSIONS to be STATIC_VERSION

seemethere:
`STATIC_VERSION` is used to denote development builds as well.

### DIFF
--- a/static/Makefile
+++ b/static/Makefile
@@ -53,7 +53,7 @@ cross-mac: cross-all-cli ## create tgz with darwin x86_64 client only
 cross-win: cross-all-cli cross-win-engine ## create zip file with windows x86_64 client and server
 	mkdir -p build/win/docker
 	cp $(CLI_DIR)/build/docker-windows-amd64 build/win/docker/docker.exe
-	cp $(ENGINE_DIR)/bundles/cross/windows/amd64/dockerd-$(VERSION).exe build/win/docker/dockerd.exe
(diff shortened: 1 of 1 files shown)


tiborvass:
I'm not sure I understand all of it, but it looks to be consistent with the rest so lgtm. But probably needs another review.

static/Makefile
```diff
@@ -53,7 +53,7 @@ cross-mac: cross-all-cli ## create tgz with darwin x86_64 client only
 cross-win: cross-all-cli cross-win-engine ## create zip file with windows x86_64 client and server
 	mkdir -p build/win/docker
 	cp $(CLI_DIR)/build/docker-windows-amd64 build/win/docker/docker.exe
-	cp $(ENGINE_DIR)/bundles/cross/windows/amd64/dockerd-$(VERSION).exe build/win/docker/dockerd.exe
+	cp $(ENGINE_DIR)/bundles/cross/windows/amd64/dockerd-$(STATIC_VERSION).exe build/win/docker/dockerd.exe
```
thaJeztah: Do we actually use `VERSION` now somewhere? If not, could we set the value to `$VERSION` directly? (`STATIC_VERSION` is a bit confusing)

seemethere: We have the same standard for Deb and rpm versioning as well since they do get transformed from the regular version.

thaJeztah: ah, I see; those use `DEB_VERSION`. Gotcha

PR #301, opened 2019-02-22
[master] [TAR-480] Add Fedora 30, make fedora builds more generic

seemethere:
Relates to https://github.com/docker/for-linux/issues/600

From the original issue (cc @jhford)
> Fedora has officially branched Version 30 today and might have a beta release on March 26.
> 
> I'm filing this issue to hopefully put this task on the radar, so that there might be a repository set up in time for the Fedora 30 release on April30
> 
> Release Schedule: https://fedoraproject.org/wiki/Releases/30/Schedule#Key_Milestones

### DIFF
--- a/rpm/fedora-30/Dockerfile
+++ b/rpm/fedora-30/Dockerfile
@@ -1,0 +1,23 @@
+ARG GO_IMAGE
+ARG ENGINE_IMAGE
+ARG BUILD_IMAGE=fedora:30
+FROM ${GO_IMAGE} as golang
+FROM ${ENGINE_IMAGE} as engine
+
+FROM ${BUILD_IMAGE}
+ENV DISTRO fedora
+ENV SUITE 30
+ENV GOPATH /go
+ENV PATH $PATH:/usr/local/go/bin:$GOPATH/bin
+ENV AUTO_GOPATH 1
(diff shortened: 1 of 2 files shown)


rpm/fedora-30/Dockerfile
```diff
@@ -0,0 +1,23 @@
+ARG GO_IMAGE
+ARG ENGINE_IMAGE
+ARG BUILD_IMAGE=fedora:30
+FROM ${GO_IMAGE} as golang
+FROM ${ENGINE_IMAGE} as engine
+
+FROM ${BUILD_IMAGE}
+ENV DISTRO fedora
+ENV SUITE 30
```
thaJeztah: Could put this in an `ARG`, then

```
ARG BUILD_IMAGE=fedora:${SUITE}
```

or

```
FROM ${BUILD_IMAGE:-fedora:${SUITE}}
```

seemethere: It could but I'd rather not have a special process specifically for `fedora` that's different from the other RPM distributions we support (both CE and EE)

thaJeztah: you'd still set the `BUILD_IMAGE` arg to specify the image, only you get rid of the double `30`

PR #308, opened 2019-03-05
Disable Kernel Memory Accounting on CentOS 7

richardwhiuk:
This applies the fix developed in https://github.com/moby/moby/pull/38128 to CentOS 7 RPMs, which are currently built without the correct flag.

This avoids kernel memory being leaked as described in https://bugzilla.redhat.com/show_bug.cgi?id=1507149

@thaJeztah you were involved in the original fix - is this something you can progress? I think this packaging change is required.

### DIFF
--- a/rpm/centos-7/Dockerfile
+++ b/rpm/centos-7/Dockerfile
@@ -11,7 +11,7 @@ ENV GOPATH=/go
 ENV PATH $PATH:/usr/local/go/bin:$GOPATH/bin
 ENV AUTO_GOPATH 1
 ENV DOCKER_BUILDTAGS pkcs11 seccomp selinux
-ENV RUNC_BUILDTAGS seccomp selinux
+ENV RUNC_BUILDTAGS seccomp selinux nokmem
 RUN yum install -y rpm-build rpmlint
 COPY SPECS /root/rpmbuild/SPECS
 # Overwrite repo that was failing on aarch64


rpm/centos-7/Dockerfile
```diff
@@ -11,7 +11,7 @@ ENV GOPATH=/go
 ENV PATH $PATH:/usr/local/go/bin:$GOPATH/bin
 ENV AUTO_GOPATH 1
 ENV DOCKER_BUILDTAGS pkcs11 seccomp selinux
-ENV RUNC_BUILDTAGS seccomp selinux
+ENV RUNC_BUILDTAGS seccomp selinux nokmem
```
thaJeztah: The `runc` version that ships with the centos7 packages is in the containerd.io package; actually wondering if this `ENV` is used at all (if it is, it would only be used for the static builds, which are not distro-specific: https://download.docker.com/linux/static/stable/x86_64/)

thaJeztah: ping @seemethere

richardwhiuk: Is the packaging for the containerd.io packages that are provided by Docker available anywhere?

thaJeztah: The repository containing the packaging scripts is not yet public; work is being done on open-sourcing those scripts, but there's still some refactoring to be done (there's some hard-coded parts in there that are specific to Docker's internal infrastructure etc.).

I think the plan is to donate it to the containerd project, and have it live somewhere in the https://github.com/containerd org at some point.

towermyeun:
Hi, sorry to but in, but does this mean `Disable kmem accounting in runc on RHEL/CentOS (docker/escalation#614, docker/escalation#692) docker/engine#121` from https://docs.docker.com/engine/#18091 is incorrect?

Is there a way to check if a particular runc binary has been compiled with the `nokmem` buildtag? Or any other way to get visibility on whether the "fix" is in-place on a system?

thaJeztah:
> Hi, sorry to butt in, but does this mean Disable kmem accounting in runc on RHEL/CentOS (docker/escalation#614, docker/escalation#692) docker/engine#121 from https://docs.docker.com/engine/#18091 is incorrect?

It's correct, but for Docker 18.09, runc is packaged separately, as part of the `containerd.io` package, and the option is set correctly in that package.

You should be able to verify by checking if the cgroup is set or not. In the case below, kernel-memory is working;

```bash
docker run --rm --kernel-memory=100M alpine cat /sys/fs/cgroup/memory/memory.kmem.limit_in_bytes
104857600
``` 

If it's disabled, you'll get an error:

```bash
docker run --rm --kernel-memory=100M alpine cat /sys/fs/cgroup/memory/memory.kmem.limit_in_bytes
WARNING: You specified a kernel memory limit on a kernel older than 4.0. Kernel memory limits are experimental on older kernels, it won't work as expected and can cause your system to be unstable.
docker: Error response from daemon: OCI runtime create failed: container_linux.go:344: starting container process caused "process_linux.go:424: container init caused \"process_linux.go:390: setting cgroup config for procHooks process caused \\\"kernel memory accounting disabled in this runc build\\\"\"": unknown.
```

seemethere:
Going to close this per @thajeztah's comment

PR #316, opened 2019-03-27
[master] Remove the pkcs11 tag, was a holdover from cli

seemethere:


tiborvass:
For reference, pkcs11 was originally added in the initial commit: https://github.com/docker/docker-ce-packaging/commit/f64fc552cdfc4b4f2efb384054970fb0bfc1bd25 and it looks like an oversight.

justincormack:
Not sure why cli has it hard coded but yeah this seems fine.

PR #317, opened 2019-04-04
bump docker app commit to baf0b24

andrewhsu:
bring it to the latest on master branch https://github.com/docker/app/tree/baf0b24

cc @vdemeester


seemethere:
This will probably need to get backported to the 19.03 branch

PR #320, opened 2019-04-09
[19.03] Add PLATFORM to deb and rpm builds

StefanScherer:
backport of #318 for 19.03.x
 Add PLATFORM to deb and rpm builds


seemethere:
Already pushed this to the 19.03 branch!

PR #325, opened 2019-04-23
[TAR-698] Add buildx as a plugin we build

seemethere:
~Depends on https://github.com/tonistiigi/buildx/pull/40~

To make sure we can consume `buildx` as part of nightly builds

### DIFF
--- a/plugins/buildx.installer
+++ b/plugins/buildx.installer
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+
+set -e
+
+source "$(dirname "$0")/.common"
+PKG=github.com/docker/buildx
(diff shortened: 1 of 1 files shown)


plugins/buildx.installer
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+
+set -e
+
+source "$(dirname "$0")/.common"
+PKG=github.com/docker/buildx
+GOPATH=$(go env GOPATH)
+REPO=https://${PKG}.git
+COMMIT=v0.2.0
```
tonistiigi: If this is for nightly builds, shouldn't it bundle nightly buildx?

PR #331, opened 2019-05-22
docker-ce-cli lacks dependency to groupadd

PhilippHomann:
I'm trying to integrate docker-ce directly to a customized fedora live-image.
Currently it fails as shadow-utils (which provides groupadd) gets installed after docker-ce-cli.

BTW:
This might have to be fixed for deb packages too.
I've noticed that in deb the docker-ce creates the docker group.
Wouldn't that be correct for rpm too?

### DIFF
--- a/rpm/SPECS/docker-ce-cli.spec
+++ b/rpm/SPECS/docker-ce-cli.spec
@@ -15,6 +15,7 @@ Packager: Docker <support@docker.com>
 
 # required packages on install
 Requires: /bin/sh
+Requires: /usr/sbin/groupadd
 
 BuildRequires: make
 BuildRequires: libtool-ltdl-devel
--- a/rpm/SPECS/docker-ce.spec
+++ b/rpm/SPECS/docker-ce.spec
@@ -15,6 +15,7 @@ URL: https://www.docker.com
 Vendor: Docker
 Packager: Docker <support@docker.com>
 
+Requires: /usr/sbin/groupadd
 Requires: docker-ce-cli
 Requires: container-selinux >= 2.9
 Requires: libseccomp >= 2.3


PhilippHomann:
Please wait until merge.
I'll also update deb packages tomorrow and move the postinstall scriptlet to docker-ce as this makes more sense imho.

PhilippHomann:
Turned out that both docker-ce and docker-ce-cli package will create the group in post-scriptlet.
Because docker-ce has dependency to docker-ce-cli, docker-ce-cli will create the group.

PhilippHomann:
@seemethere would you like to approve the fix?

viliampucik:
Hi @PhilippHomann and @seemethere :) Please, is there any progress on this?

Also I assume the correct pull request should contain `Requires(post)` instead of `Requires`:

```
Requires(post):  /usr/sbin/groupadd
```

thaJeztah:
Curious; does ` Depends: /usr/sbin/groupadd` actually manage installing the dependency? (given that it's not a package name)?

PhilippHomann:
Yes, it does because it requires the file to be installed, no matter by whom.
And as the file list is known by a package manager like yum, it installs a package providing that file.

viliampucik:
@thaJeztah `Depends` keyword does not seem to be present in the specification (https://rpm-packaging-guide.github.io/).

viliampucik:
Noting this PR also fixes the installation issue in CentOS 8.

thaJeztah:
> Yes, it does because it requires the file to be installed, no matter by whom. And as the file list is known by a package manager like yum, it installs a package providing that file.

Thanks! I know that the `Requires: /bin/sh` was in there, and that always stood a bit out, but never dived into it 😅 

> EDIT: as correctly stated below, it is requires, not depends.

😊 whoops; not near my computer, and mis-typed on my phone

PR #332, opened 2019-05-23
[WIP] Support for experimental plugins in packaging

tiborvass:

To be tested.

Related to https://github.com/docker/cli/pull/1902

### DIFF
--- a/plugins/.common
+++ b/plugins/.common
@@ -2,6 +2,7 @@
 
 DESTDIR=${DESTDIR:-}
 PREFIX=${PREFIX:-/usr/local}
+EXPERIMENTAL_PLUGIN=${EXPERIMENTAL_PLUGIN:-}
 
 add_github_ssh_host() {
     # You're not able to clone from github unless you add to known_hosts
@@ -12,6 +13,7 @@ add_github_ssh_host() {
 }
 
 install_binary() {
+    [ -n "${EXPERIMENTAL_PLUGIN}" ] && PREFIX=${PREFIX}-experimental
     for binary in "$@"; do
         mkdir -p "${DESTDIR}${PREFIX}"
         install -p -m 755 "${binary}" "${DESTDIR}${PREFIX}"
--- a/plugins/buildx.installer
+++ b/plugins/buildx.installer
@@ -8,6 +8,7 @@ GOPATH=$(go env GOPATH)
 REPO=https://${PKG}.git
 COMMIT=master
 DEST=${GOPATH}/src/${PKG}
+export EXPERIMENTAL_PLUGIN=1
 
 build() {
     if [ ! -d "${DEST}" ]; then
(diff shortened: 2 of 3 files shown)


plugins/buildx.installer
```diff
@@ -18,8 +19,10 @@ build() {
         git fetch --all
         git checkout -q "${COMMIT}"
         local LDFLAGS
-        # TODO: unmark `-tp` when no longer a technical preview
-        LDFLAGS="-X ${PKG}/version.Version=$(git describe --match 'v[0-9]*' --always --tags)-tp -X ${PKG}/version.Revision=$(git rev-parse HEAD) -X ${PKG}/version.Package=${PKG}"
+        local VERSION
+        VERSION=$(git describe --match 'v[0-9]*' --always --tags)
+        test -n "${EXPERIMENTAL_PLUGIN}" && VERSION=${VERSION}-tp
```
seemethere: Wait what is this supposed to do?

tiborvass: I can remove it. It just does what you had in a comment: add `-tp` if it's experimental.

seemethere:
Merge is currently pending https://github.com/docker/cli/pull/1902

tiborvass:
Similar will be needed in ee-packaging for ee-specific plugins that need to be experimental.

plugins/.common
```diff
@@ -12,6 +13,7 @@ add_github_ssh_host() {
 }
 
 install_binary() {
+    [ -n "${EXPERIMENTAL_PLUGIN}" ] && PREFIX=${PREFIX}-experimental
```
seemethere: Might be preferable to do something like this:

```suggestion
	local install_prefix="${PREFIX}"
    if [ -n "${EXPERIMENTAL_PLUGIN}" ]; then
    	install_prefix="${PREFIX}-experimental"
   	fi
```

Then change line 19 from `${PREFIX}` to `${install_prefix}` that way we can avoid changing global variables

plugins/.common
```diff
@@ -12,6 +13,7 @@ add_github_ssh_host() {
...
     for binary in "$@"; do
         mkdir -p "${DESTDIR}${PREFIX}"
         install -p -m 755 "${binary}" "${DESTDIR}${PREFIX}"
```
seemethere:
```suggestion
        install -p -m 755 "${binary}" "${DESTDIR}${install_prefix}"
```

plugins/buildx.installer
```diff
@@ -18,8 +19,10 @@ build() {
         git fetch --all
         git checkout -q "${COMMIT}"
         local LDFLAGS
-        # TODO: unmark `-tp` when no longer a technical preview
-        LDFLAGS="-X ${PKG}/version.Version=$(git describe --match 'v[0-9]*' --always --tags)-tp -X ${PKG}/version.Revision=$(git rev-parse HEAD) -X ${PKG}/version.Package=${PKG}"
+        local VERSION
```
seemethere: Might also be useful to name this something different as to avoid overwriting the global `VERSION` we use for the dockerd/docker-cli builds.

seemethere:
I think this one can be closed

PR #343, opened 2019-06-17
TAR-875 Fix man-pages showing "minimized" message

thaJeztah:
The dh_installman script calls "man" and captures its output to convert manpages
to utf8:

https://github.com/Debian/debhelper/blob/8523120dccaf5666425109da228b7e1778f15e8b/dh_installman#L298-L316

however, on minimized Ubuntu systems, man is overridden by a script that outputs
a warning message ("This  system  has been minimized by removing packages and
content ..").

As a result, all man-pages were be overwritten by that message.

This patch restores the actual `man` command before running `dh_installman` to
work around this issue.

addresses docker/for-linux#639

### DIFF
--- a/deb/common/rules
+++ b/deb/common/rules
@@ -63,5 +63,13 @@ override_dh_install:
 	# TODO Can we do this from within our container?
 	dh_apparmor --profile-name=docker-ce -pdocker-ce
 
+override_dh_installman:
+	# based on ubuntu's "unminimize" script
+	if [ "$$(dpkg-divert --truename /usr/bin/man)" = "/usr/bin/man.REAL" ]; then \
+		rm -f /usr/bin/man \
+		&& dpkg-divert --quiet --remove --rename /usr/bin/man; \
+	fi
+	dh_installman
+
 %:
 	dh $@ --with=bash-completion $(shell command -v dh_systemd_enable > /dev/null 2>&1 && echo --with=systemd)


deb/common/rules
```diff
@@ -63,5 +63,13 @@ override_dh_install:
 	# TODO Can we do this from within our container?
 	dh_apparmor --profile-name=docker-ce -pdocker-ce
 
+override_dh_installman:
+	# based on ubuntu's "unminimize" script
+	if [ "$$(dpkg-divert --truename /usr/bin/man)" = "/usr/bin/man.REAL" ]; then \
+		rm -f /usr/bin/man \
+		&& dpkg-divert --quiet --remove --rename /usr/bin/man; \
```
thaJeztah: took this from the Ubuntu script, but perhaps a simple `mv /usr/bin/man.REAL /usr/bin/man` would be enough

thaJeztah:
ping @kolyshkin @seemethere PTAL

deb/common/rules
seemethere: I'm just wondering if this is actually the correct to way to do this, like should our package be the one to "uniminimize" man pages or should that be handled by ubuntu's own unminimizing scripts.

Like the OS had a reason to minimize so I don't think that we should be the one's to go through and unminimize on behalf of the OS.

thaJeztah: this would only replace the `man` script with the actual `man` binary; and only during our packaging. Not sure what other options we have; we should report a bug upstream I guess (so that the `dh_installman` script fails if it's not an actual `man`, but the upstream is Debian, so I guess they would point to Ubuntu

rgson: I have filed a bug report against Ubuntu's version of the debhelper package: <https://bugs.launchpad.net/ubuntu/+source/debhelper/+bug/1833124>

thaJeztah:
@seemethere LMK if you want a different approach; I agree that (ideally), building a `.deb` shouldn't modify the state of the machine; in our case "machine" is the container in which the build is run so I guess it's "ok(ish)" for our specific use; it's something to be fixed upstream though (imo). (Also, the script _unconditionally_ converting to utf8 without first checking if conversion is needed is a bit odd)

kolyshkin:
I agree that changing the build env from debian/rules is not good. This fix belongs elsewhere (to whatever creates that build environment).

In here, we can just check that the build environment is correct -- if not, bail out and provide some explanation/instructions.

thaJeztah:
Alternative approach in https://github.com/docker/docker-ce-packaging/pull/350

PR #345, opened 2019-06-25
require container-selinux >= 2.74

thaJeztah:
version 2.9 is really old; this sets the same minimal version as is used for the containerd.io package

### DIFF
--- a/rpm/SPECS/docker-ce.spec
+++ b/rpm/SPECS/docker-ce.spec
@@ -16,7 +16,7 @@ Vendor: Docker
 Packager: Docker <support@docker.com>
 
 Requires: docker-ce-cli
-Requires: container-selinux >= 2.9
+Requires: container-selinux >= 2:2.74
 Requires: libseccomp >= 2.3
 Requires: systemd-units
 Requires: iptables


thaJeztah:
ping @andrewhsu @kolyshkin @justincormack PTAL

This is the same version as we set in the containerd.io package, but the discussion in https://github.com/containers/container-selinux/issues/63 goes about version 2.99; should we set the minimal version higher than this (to get the fix from https://github.com/containers/container-selinux/pull/64 in?)

Looks like 2.99 is available on CentOS now;

```
==================================================================================================================================================================================================================================================================================================================================================
 Package                                                                                     Arch                                                                       Version                                                                                 Repository                                                                   Size
==================================================================================================================================================================================================================================================================================================================================================
Installing:
 container-selinux                                                                           noarch                                                                     2:2.99-1.el7_6                                                                          extras                                                                       39 k
Installing for dependencies:
```

thaJeztah:
oh actually looks like we changed `runc` back to `/usr/bin`, so I guess we don't require `v2.99`

kolyshkin:
At this point I do not remember which version is good and which one is not, but I think that 2.95 is OK and 2.74 might be problematic. @seemethere might remember more

rpm/SPECS/docker-ce.spec
```diff
@@ -16,7 +16,7 @@ Vendor: Docker
 Packager: Docker <support@docker.com>
 
 Requires: docker-ce-cli
-Requires: container-selinux >= 2.9
+Requires: container-selinux >= 2:2.74
```
kolyshkin: If we're to build this for openSUSE/SLES, we need this:
https://github.com/docker/containerd-packaging/pull/95/commits/072a3aad8bf4c593eceabf3a3f11658792208bdb

kolyshkin:
OK based on what I see in 
https://github.com/docker/containerd-packaging/pull/95 I think `>=2.74` is good

rpm/SPECS/docker-ce.spec
thaJeztah: these specs are only used for Docker CE, there's a spec in the EE packaging repository that does that

PR #347, opened 2019-06-25
Sync RPM Spec with downstream EE packaging

thaJeztah:
This takes some of the changes from the downstream EE packaging scripts to bring the specs more in line.

- systemd-units -> systemd (taken from `8bb1f0a7a395dfd979cd410b26ef47e55433de32`)
- some wording changes (taken from `a8c522a7c22c34dc354e91941377a4aadc3ddc4c`)

Remove the pre and posttrans, they weren't useful

On some distros we were encountering errors where `$1` was not being populated
for the 'pre' and 'posttrans' rpm macros, upon closer inspection it
isn't exactly clear why the pre and posttrans macro scripts were exactly
useful since the `%systemd_postun_with_restart` does exactly what those
scripts were doing.

I've tidied up the systemd macros to use `docker.service` instead of
`docker` which seems to function as we'd expect.

taken from downstream commit `62d8413b550659a0b5318346ee2e3d7e4a50bfe1`

### DIFF
--- a/rpm/SPECS/docker-ce.spec
+++ b/rpm/SPECS/docker-ce.spec
@@ -18,7 +18,7 @@ Packager: Docker <support@docker.com>
 Requires: docker-ce-cli
 Requires: container-selinux >= 2.9
 Requires: libseccomp >= 2.3
-Requires: systemd-units
+Requires: systemd
 Requires: iptables
 Requires: libcgroup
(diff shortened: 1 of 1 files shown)


thaJeztah:
ping @andrewhsu @StefanScherer @seemethere PTAL

PR #349, opened 2019-06-25
Add Raspbian buster

ndanyluk:
This PR adds a docker file for raspbian buster and modifies the makefile to include it.  I followed the same format in the makefile as was used for raspbian-stretch and debian-buster


thaJeztah:
thanks! will have to give this a look, but could you squash the first two commits? (second one looks like it's fixing up a thing from the first one)

let me know if you need help doing so

thaJeztah:
also: @seemethere @zelahi PTAL

ndanyluk:
Ok I think I got the commits squashed properly.  I've never done that before, but I think it worked.  Please let me know if there's any other changes you want me to make and I'd be happy to help! @thaJeztah @seemethere

PR #357, opened 2019-07-01
Assorted improvements to Makefiles

kolyshkin:
## Simplify and fix rpm/Makefile and deb/Makefile

1. A symbol `%` is not working as expected in `PHONY` targets, so
e.g. `fedora-30` or `ubuntu-cosmic` were not being rebuild each time,
because `fedora-30` and `ubuntu-cosmic` directories exists.
The fix is to explicitly mark every distro target as phony.
This also has the nice side effect of bash completion
working as it should!
    
2. Remove code duplication for making packages for different distros.

3. Add missing ubuntu releases (cosmic and disco).

## Makefile: rely on targets in deb/rpm
    
Instead of dynamically getting list of distros to build for,
rely on the corresponding targets in sub-Makefiles. This also
ensures that deb/Makefile and rpm/Makefile will have up-to-date
list of distros included.
    
This also fixes the following bug:
    
> $ make deb
> for p in raspbian-stretch ubuntu-bionic ubuntu-disco ubuntu-xenial debbuild/ubuntu-disco ubuntu-cosmic debian-buster debian-stretch; do \
> ...
    
As you can see, `debbuild/ubuntu-disco` should not be included but it
is. Could be prevented by using `-maxdepth 1` argument to `find`.
    
While at it, amend the sub-Makefiles to print out the distro
that we build for.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -29,7 +29,7 @@ test_steps = [
 				sh("git -C cli checkout $branch")
 				sh('git clone https://github.com/docker/engine.git')
 				sh("git -C engine checkout $branch")
-				sh('make VERSION=0.0.1-dev DOCKER_BUILD_PKGS=ubuntu-xenial ENGINE_DIR=$(pwd)/engine CLI_DIR=$(pwd)/cli deb')
+				sh('make -C deb VERSION=0.0.1-dev ENGINE_DIR=$(pwd)/engine CLI_DIR=$(pwd)/cli ubuntu-xenial')
 			}
 		}
 	},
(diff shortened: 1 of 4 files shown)


thaJeztah:
looks like this needs a rebase

thaJeztah:
Guess we'll have to backport this to the release branches (once merged)

kolyshkin:
* rebased to HEAD (added raspbian-buster)
* fixed to not build raspbian by default from the main Makefile
* fixed Jenkinsfile to work with updated Makefiles

kolyshkin:
Can't figure out the CI failure, what do I need to look for?

thaJeztah:
@kolyshkin looks like this is the failure; https://jenkins.dockerproject.org/job/docker/job/docker-ce-packaging/job/PR-357/3/execution/node/160/log/

```
21:45:12 Makefile:76: *** missing separator (did you mean TAB instead of 8 spaces?).  Stop.
```

kolyshkin:
ah, stupid copy-paste error ... my `vim` used to mark those but it isn't now:(

> 00:00:48.834 Makefile:76: *** missing separator (did you mean TAB instead of 8 spaces?).  Stop.

Fixed, pushed

PR #369, opened 2019-08-06
Pin Golang images to debian stretch variant

thaJeztah:
the default switched to debian buster, which will have side-effects (e.g. `libgcc-6-dev` is no longer in buster)


thaJeztah:
ping @seemethere @kolyshkin PTAL

thaJeztah:
Actually; not sure if we need this, because we only copy the Go binaries 🤔

thaJeztah:
oh, never mind; we _do_ use the golang image for the image variant;

https://github.com/docker/docker-ce-packaging/blob/06ff8d13622bd0fd9360e7d5c628e8d2bf8bb75a/image/Dockerfile.engine#L3-L15

thaJeztah:
yeah, there's more side effects to switching to buster; see https://github.com/moby/moby/pull/39549

PR #373, opened 2019-08-14
static: only set GO_VERSION if it's empty

thaJeztah:
I think this only has to be done if no `GO_VERSION` was passed / if the version is empty


thaJeztah:
@zelahi @seemethere PTAL - trying to figure out how all the makefiles hook into each other 😂

PR #375, opened 2019-08-14
Makefile: use an include file for common variables

thaJeztah:
With the assumption that;

- each makefile should be able to be used standalone (`cd static && make <foo>`)
- each makefile should also be able to be called from the top makefile (`make static` -> recurses over the makefiles)
- when called from the main makefile, common variables that are set in the main makefile should be (mostly) leading.
- it should be possible to manually override variables (`make GO_VERSION=1.13beta1 static`)

Tricky bits:

- paths ... because some Makefiles are in subdirectories, we cannot set (e.g.) `CLI_DIR` and `ENGINE_DIR` in the `common.mk` (because it's done relative to the _working_ directory).
  - Perhaps some trickery could be done with `realpath` (docs says it returns an empty path if it's not found, so we could first try `../engine` and if it's empty, try `../../engine`)
- `GO_VERSION`; this is a bit unclear: why does `static/Makefile` detect the Go-version from the CLI, but other targets don't?
  - Should `GO_VERSION` be set to the CLI's version for all targets?
  - Or set to `latest` ?
  - Or empty (force a version to be set?)

`STATIC_VERSION` (see first commit)

I noticed that static version was always overridden (?), which meant that static packages would potentially get a different version than non-static versions if called from the main Makefile (`make static`)? (TBH; just from reading the code, haven't verified if they did)


thaJeztah:
@seemethere @zelahi

PR #393, opened 2019-10-09
Jenkinsfile: change to declarative with params

andrewhsu:
This PR changes the `Jenkinsfile` from scripted pipeline format to the new declarative pipeline format that is preferred by Jenkins. Notable changes:
* runs all the same package tests except the aarch64 and armhf
* parameters for specifying source to git clone the engine and cli
* parameter to archive packages after built, default: false

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,72 +1,123 @@
 #!groovy
-
-
-def genBranch(String arch) {
-	return [
-		"${arch}": { ->
-			stage("Build engine image on ${arch}") {
-				wrappedNode(label: "linux&&${arch}", cleanWorkspace: true) {
-					try {
-						checkout scm
-						sh("git clone https://github.com/docker/engine.git engine")
-						sh('make ENGINE_DIR=$(pwd)/engine image')
-					} finally {
-						sh('make ENGINE_DIR=$(pwd)/engine clean-image clean-engine')
+pipeline {
+	agent none
+	options {
+		buildDiscarder(logRotator(daysToKeepStr: '30'))
+		timeout(time: 30, unit: 'MINUTES')
+		timestamps()
+	}
+	parameters {
+		string(name: 'github_repo_engine', defaultValue: 'docker/engine', description: 'github org/repo of engine')
+		string(name: 'github_branch_engine', defaultValue: env.CHANGE_TARGET, description: 'github branch of engine')
+		string(name: 'github_repo_cli', defaultValue: 'docker/cli', description: 'github org/repo of cli')
+		string(name: 'github_branch_cli', defaultValue: env.CHANGE_TARGET, description: 'github branch of cli')
+		booleanParam(name: 'archive_packages', defaultValue: false, description: 'archive packages')
+	}
+	environment {
+		DOCKER_BUILDKIT = '1'
+	}
+	stages {
+		stage('Build') {
+			parallel {
+				stage('ubuntu-xenial') {
+					agent { label 'amd64 && ubuntu-1804 && overlay2' }
+					stages {
+						stage('create package') {
+							steps {
+								sh 'docker version'
+								sh 'docker info'
+								sh "git clone -b '$params.github_branch_engine' --depth 1 'https://github.com/$params.github_repo_engine' engine"
+								sh "git clone -b '$params.github_branch_cli' --depth 1 'https://github.com/$params.github_repo_cli' cli"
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -1,72 +1,123 @@
...
+			parallel {
+				stage('ubuntu-xenial') {
+					agent { label 'amd64 && ubuntu-1804 && overlay2' }
+					stages {
+						stage('create package') {
+							steps {
+								sh 'docker version'
+								sh 'docker info'
+								sh "git clone -b '$params.github_branch_engine' --depth 1 'https://github.com/$params.github_repo_engine' engine"
+								sh "git clone -b '$params.github_branch_cli' --depth 1 'https://github.com/$params.github_repo_cli' cli"
```
seemethere: If we're doing this on every stage then we might as well just make this a function.

Jenkinsfile
```diff
@@ -1,72 +1,123 @@
...
+							when { expression { params.archive_packages } }
+							steps {
+								archiveArtifacts artifacts: 'deb/debbuild/ubuntu-xenial/docker-ce*.deb'
+							}
+						}
+					}
+					post {
+						cleanup {
+							sh 'docker run --rm -v "$WORKSPACE:/workspace" busybox chown -R "$(id -u):$(id -g)" /workspace'
+							deleteDir()
```
seemethere: We could also make this a function as well.

Jenkinsfile
thaJeztah: Don't think declarative pipelines (officially) allow creating functions

Could be a make target though, and we could use env-vars to pass this information to them, which could be set as a global env-var for all stages

Jenkinsfile
```diff
@@ -1,72 +1,123 @@
...
+								sh 'docker info'
+								sh "git clone -b '$params.github_branch_engine' --depth 1 'https://github.com/$params.github_repo_engine' engine"
+								sh "git clone -b '$params.github_branch_cli' --depth 1 'https://github.com/$params.github_repo_cli' cli"
+								sh 'make -C deb VERSION=0.0.1-dev ENGINE_DIR=$(pwd)/engine CLI_DIR=$(pwd)/cli ubuntu-xenial'
+							}
+						}
+						stage('archive package') {
+							when { expression { params.archive_packages } }
+							steps {
+								archiveArtifacts artifacts: 'deb/debbuild/ubuntu-xenial/docker-ce*.deb'
```
thaJeztah: Could this use a wildcard for the distro?

```suggestion
								archiveArtifacts artifacts: 'deb/debbuild/**/docker-ce*.deb'
```

Jenkinsfile
```diff
@@ -1,72 +1,123 @@
...
+				stage('ubuntu-xenial') {
+					agent { label 'amd64 && ubuntu-1804 && overlay2' }
+					stages {
+						stage('create package') {
+							steps {
+								sh 'docker version'
+								sh 'docker info'
+								sh "git clone -b '$params.github_branch_engine' --depth 1 'https://github.com/$params.github_repo_engine' engine"
+								sh "git clone -b '$params.github_branch_cli' --depth 1 'https://github.com/$params.github_repo_cli' cli"
+								sh 'make -C deb VERSION=0.0.1-dev ENGINE_DIR=$(pwd)/engine CLI_DIR=$(pwd)/cli ubuntu-xenial'
```
thaJeztah: Should we have a `make install` / `make verify` step that

- installs the package that was built
- prints the output of `docker version`, `docker info`
- runs a (some) container(s)

Jenkinsfile
```diff
@@ -1,72 +1,123 @@
...
+		buildDiscarder(logRotator(daysToKeepStr: '30'))
+		timeout(time: 30, unit: 'MINUTES')
+		timestamps()
+	}
+	parameters {
+		string(name: 'github_repo_engine', defaultValue: 'docker/engine', description: 'github org/repo of engine')
+		string(name: 'github_branch_engine', defaultValue: env.CHANGE_TARGET ?: env.BRANCH_NAME, description: 'github branch of engine')
+		string(name: 'github_repo_cli', defaultValue: 'docker/cli', description: 'github org/repo of cli')
+		string(name: 'github_branch_cli', defaultValue: env.CHANGE_TARGET ?: env.BRANCH_NAME, description: 'github branch of cli')
+		booleanParam(name: 'archive_packages', defaultValue: false, description: 'archive packages')
```
seemethere: I don't think we need this, let's just archive everything by default

PR #397, opened 2019-10-17
fix eq issue

daixiang0:
In linux, this check always return 0:
```
$ [[ $(uname) -eq "Darwin" ]] && echo yes
yes
$ uname
Linux
```

Signed-off-by: Xiang Dai <764524258@qq.com>


daixiang0:
@droopy4096 better to backport to other branch?

PR #404, opened 2019-10-25
deb: Only recommend aufs-tools on amd64

seemethere:
To remedy issues found for users not running `amd64`, ex: https://github.com/raspberrypi/linux/issues/3021


thaJeztah:
Even wondering if we should remove the recommends altogether, but I don't think there's anything less than "recommend", correct?

seemethere:
Yeah I believe recommends is the lowest level of dependency declaration we can make

PR #409, opened 2019-11-08
[TAR-1335][19.03] Add optional dependency on nftables for RHEL8+ derivatives

droopy4096:


zelahi:
@droopy4096 should this be opened against master: https://github.com/docker/docker-ce-packaging/blob/master/rpm/SPECS/docker-ce.spec and then we backport it to this branch.  It would help when we are dealing with merge conflicts

zelahi:
Another small knitpit.  Sorry I modified the title a bit.  I added the 19.03 as a mini heading.  It's so we an tell what branch this is being opened against =)

droopy4096:
@zelahi now we have branch name in two places ;) I'll remove the suffix one...

rpm/SPECS/docker-ce.spec
```diff
@@ -19,7 +19,11 @@ Requires: docker-ce-cli
 Requires: container-selinux >= 2:2.74
 Requires: libseccomp >= 2.3
 Requires: systemd
+%if 0%{?rhel} >= 8
+Requires: iptables or nftables
```
thaJeztah: I'm curious; I would've expected the `nftables` package to "provide" `iptables`, is that not the case?

droopy4096: I'll file the second PR for the master.

nope - nftables does *not* provide `iptables` 
```shell
$ rpm --provides -q nftables
config(nftables) = 1:0.9.0-8.el8
libnftables.so.0()(64bit)
nftables = 1:0.9.0-8.el8
nftables(x86-64) = 1:0.9.0-8.el8
```

thaJeztah:
@droopy4096 I see you closed this one; are you working on a PR for master?

droopy4096:
@thaJeztah  PR-414

PR #410, opened 2019-11-28
RHEL8 build fix

droopy4096:
fix SPEC file declaration for boolean dependency


dave-tucker:
Ignore me @droopy4096. Not sure squashing is right at all.
I think the cherry pick of `a674309` might have gone wrong.
Did you use `git cherry-pick -x`?

droopy4096:
squashed before merge :)

PR #411, opened 2019-12-05
Remove image-based build (for docker engine activate)

thaJeztah:
related to https://github.com/docker/cli/pull/2207
related to https://github.com/docker/docker-ce-packaging/pull/416

closes https://github.com/docker/docker-ce-packaging/pull/417

PR #417, opened 2020-01-17
Fix image based build to address for upstream changes

thaJeztah:
depends on https://github.com/moby/moby/pull/40393

Upstream moby/moby#40180 removed the auto-generated code, replacing
it with build-time variables (-X).

This patch updates the Dockerfiles to account for this change.

### DIFF
--- a/image/Dockerfile.engine
+++ b/image/Dockerfile.engine
@@ -31,19 +31,22 @@ ARG BUILDTIME
 ARG PLATFORM
 ARG PRODUCT
 ARG DEFAULT_PRODUCT_LICENSE
+ARG TINI_COMMIT
 ENV VERSION ${VERSION}
 ENV GITCOMMIT ${GITCOMMIT}
 ENV BUILDTIME ${BUILDTIME}
 ENV PLATFORM ${PLATFORM}
 ENV PRODUCT ${PRODUCT}
 ENV DEFAULT_PRODUCT_LICENSE ${DEFAULT_PRODUCT_LICENSE}
+ENV TINI_COMMIT ${TINI_COMMIT}
+SHELL ["/bin/bash", "-c"]
 # TODO The way we set the version could easily be simplified not to depend on hack/...
-RUN bash ./hack/make/.go-autogen
-RUN go build -o /sbin/dockerd \
-    -tags 'autogen apparmor seccomp selinux journald exclude_graphdriver_devicemapper' \
+RUN source ./hack/make/.go-autogen; \
+    go build -o /sbin/dockerd \
+    -tags 'apparmor seccomp selinux journald exclude_graphdriver_devicemapper' \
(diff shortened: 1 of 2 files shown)


thaJeztah:
ping @arkodg @zelahi @StefanScherer PTAL 🤗

image/Dockerfile.engine-dm
```diff
@@ -37,19 +37,22 @@ ARG BUILDTIME
...
 ENV PRODUCT ${PRODUCT}
 ENV DEFAULT_PRODUCT_LICENSE ${DEFAULT_PRODUCT_LICENSE}
+ENV TINI_COMMIT ${TINI_COMMIT}
+SHELL ["/bin/bash", "-c"]
 # TODO The way we set the version could easily be simplified not to depend on hack/...
-RUN bash ./hack/make/.go-autogen
-RUN go build -o /dockerd \
-    -tags 'autogen apparmor seccomp selinux journald' \
+RUN source ./hack/make/.go-autogen; \
+    go build -o /sbin/dockerd \
```
StefanScherer: This `/sbin/dockerd` will cause a problem in the last stage. There's a `COPY --from` in line 80 that uses `/dockerd` as source.

This lead to a next question, why are there two relatively similar `Dockerfile.engine*` files that we have to maintain?

At least we should use the same pathes in both Dockerfiles, so the COPY instruction must be updated.

image/Dockerfile.engine
```diff
@@ -31,19 +31,22 @@ ARG BUILDTIME
 ARG PLATFORM
 ARG PRODUCT
 ARG DEFAULT_PRODUCT_LICENSE
+ARG TINI_COMMIT
```
StefanScherer: Do we need to pass in that argument somewhere with `--build-arg TINI_COMMIT` ?

image/Dockerfile.engine-dm
thaJeztah: 🤦‍♂ I worked on one file, then copied to the other file, but forgot they had different paths.

image/Dockerfile.engine
thaJeztah: Hm.. good one; let me check if it's passed somewhere

thaJeztah: even wondering how this worked at all, because the `binaries-commits` file was removed in https://github.com/moby/moby/commit/b529d1b0936b90ae14d584c73f7332919f8d76b7

thaJeztah:
Yeah, perhaps we should just go for https://github.com/docker/docker-ce-packaging/pull/411, which removes this stuff. I started with this because I thought the problem was in build in general, but it was only in the image-based build.

thaJeztah:
superseded by https://github.com/docker/docker-ce-packaging/pull/411

PR #424, opened 2020-01-21
[19.03] Jenkinsfile: fix image-based engine using wrong branch

thaJeztah:
This caused CI to build the engine from master, which has some changes that makes building on this branch fail.

this should help https://github.com/docker/docker-ce-packaging/pull/419 pass

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,14 +1,15 @@
 #!groovy
 
 
-def genBranch(String arch) {
+def genBranch(String arch, String branch) {
 	return [
 		"${arch}": { ->
 			stage("Build engine image on ${arch}") {
-				wrappedNode(label: "linux&&${arch}", cleanWorkspace: true) {
+				wrappedNode(label: "linux && ${arch}", cleanWorkspace: true) {
 					try {
 						checkout scm
-						sh("git clone https://github.com/docker/engine.git engine")
+						sh("git clone https://github.com/docker/engine.git")
+						sh("git -C engine checkout $branch")
 						sh('make ENGINE_DIR=$(pwd)/engine image')
 					} finally {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -8,7 +8,8 @@ def genBranch(String arch) {
 				wrappedNode(label: "linux&&${arch}", cleanWorkspace: true) {
 					try {
 						checkout scm
-						sh("git clone https://github.com/docker/engine.git engine")
+						sh("git clone https://github.com/docker/engine.git")
+						sh("git -C engine checkout $branch")
```
thaJeztah: For some reason, it only does the `git clone`, and after that continues with the `finally` (`make ENGINE_DIR=$(pwd)/engine clean-image clean-engine`).

So for some reason, the `git clone` is failing, but no error is shown. Perhaps a permissions issue, or the `engine` directory exists (and is not empty?)

thaJeztah:
Added some debug lines, but no clue why it's going to the `finally` directly 😞 

Screenshot 2020-01-22 at 09 28 05

thaJeztah:
ooooh, I think this might be it; the actual error is printed at the end of the output of the `finally` branch;

```
No such property: branch for class: groovy.lang.Binding
```

So `$branch` is not defined there, which could make sense?

thaJeztah:
ping @StefanScherer @silvin-lubecki PTAL

thaJeztah:
Perhaps for 18.09 (but that's EOL). This is already for the 19.03 branch, and this part was removed on master (and may be removed in 19.03 as well at some point, but I didn't want to make the changes too big)

thaJeztah:
Thanks for reviewing; bringing this one in

PR #434, opened 2020-02-24
[master] Check built packages

silvin-lubecki:
Added scripts to check `deb` and `rpm` packages just after building them.

image

### DIFF
--- a/deb/Makefile
+++ b/deb/Makefile
@@ -4,17 +4,16 @@ CLI_DIR:=$(realpath $(CURDIR)/../../cli)
 ENGINE_DIR:=$(realpath $(CURDIR)/../../engine)
 PLUGINS_DIR:=$(realpath $(CURDIR)/../plugins)
 GITCOMMIT?=$(shell cd $(CLI_DIR) && git rev-parse --short HEAD)
-STATIC_VERSION:=$(shell ../static/gen-static-ver $(ENGINE_DIR) $(VERSION))
 GO_BASE_IMAGE=golang
 GO_IMAGE=$(GO_BASE_IMAGE):$(GO_VERSION)-stretch
 DEB_VERSION=$(shell ./gen-deb-ver $(CLI_DIR) "$(VERSION)")
-CHOWN:=docker run --rm -v $(CURDIR):/v -w /v alpine chown
 EPOCH?=5
 
 ifdef BUILD_IMAGE
 	BUILD_IMAGE_FLAG=--build-arg $(BUILD_IMAGE)
 endif
 
+# Build docker image used to build packages
 COMMON_FILES=common
 BUILD?=DOCKER_BUILDKIT=1 \
 	docker build \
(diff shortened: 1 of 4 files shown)


deb/verify
```diff
@@ -0,0 +1,81 @@
...
+        # 'apt-get' would attempt to install dependency packages (such as the CLI)
+        # from download.docker.com instead of the locally built CLI package. Given
+        # that 'dpkg -i' does not install any dependency (but will fail if depen-
+        # dencies are missing), we use the '--ignore-depends' option to ignore
+        # packages we know to be missing at this stage, and '--force-depends' to
+        # only warn about any other missing dependency.
+        #
+        # Afterwards, we run 'apt-get install' with the '--fix-broken' option to
+        # trigger installation of the dependencies, which should succceed succesfully.
+        dpkg \
```
seemethere: You can also achieve this with just a simple `apt-get install ./${packages}`, that's worked to install local packges for a while now while still using the same `apt-get` resolver

thaJeztah: ISTR there was _something_ that didn't work, but don't recall what. Does it also support the `--ignore-depends` and `--force-depends` options?

thaJeztah:
@silvin-lubecki there's some fixes that weren't merged yet in the downstream repo, so we may want to wait for those to be included first

deb/verify
seemethere: It should be compatible with all flags, idk why apt doesn't just work with local files without adding the prefixed `./` but it should *just work*.

This is the verification model I was using for `containerd.io` before leaving.

silvin-lubecki:
PTAL @thaJeztah  🙏

deb/verify
silvin-lubecki: Thanks @seemethere ! I'll try that in a follow-up as this PR is mainly moving code 👍

rpm/verify
```diff
@@ -0,0 +1,82 @@
...
+	# install all non-source packages
+	(
+		set -x
+		${pkg_manager} install -y ${packages}
+	)
+
+	# Verify docker version is the one expected.
+	expected_ver=$1
+	expected_gitcommit=$2
+	docker_ver=`docker --version`
```
thaJeztah: We should probably do this twice; once for `docker`, and once for `dockerd` (and/or use `docker version --format ....`)

deb/Makefile
```diff
@@ -4,17 +4,16 @@ CLI_DIR:=$(realpath $(CURDIR)/../../cli)
 ENGINE_DIR:=$(realpath $(CURDIR)/../../engine)
 PLUGINS_DIR:=$(realpath $(CURDIR)/../plugins)
 GITCOMMIT?=$(shell cd $(CLI_DIR) && git rev-parse --short HEAD)
-STATIC_VERSION:=$(shell ../static/gen-static-ver $(ENGINE_DIR) $(VERSION))
```
thaJeztah: Does the `static` version also have a check?

deb/verify
```diff
@@ -0,0 +1,89 @@
...
+		else
+			echo "FAILED! docker version does not match expected $expected_ver"
+			exit 1
+		fi
+		if [[ $installed_gitcommit =~ ^$expected_gitcommit ]]; then
+			echo "PASSED! docker git commit matches expected $expected_gitcommit"
+		else
+			echo "FAILED! docker git commit does not match expected $expected_gitcommit"
+			exit 2
+		fi
```
thaJeztah: I think this check is not specific to deb/rpm/static, so this could be considered to be put in a function + common file, and shared between them

deb/Makefile
silvin-lubecki: No, but I can add one 👍

deb/verify
silvin-lubecki: Right, I'll find a common place 👍

silvin-lubecki:
Closing this one for now

PR #441, opened 2020-03-04
Don't disable socket activation when stopping docker service

thaJeztah:
PartOf deactivates the socket whenever the service get deactivated.
The socket unit however should be active nevertheless, so that the
docker service can be started again through socket activation.

Based on the original patch in upstream moby/moby by Max Harmathy (https://github.com/moby/moby/pull/37470).

Fixes https://github.com/moby/moby/issues/37468


silvin-lubecki:
That makes me think we lack e2e tests... 🤔

PR #444, opened 2020-03-20
Fix CentOS 8 build

thaJeztah:
- Enable PowerTools repo
- Disable building btrfs, which is no longer supported on
  CentOS/RHEL

### DIFF
--- a/rpm/SPECS/docker-ce.spec
+++ b/rpm/SPECS/docker-ce.spec
@@ -1,5 +1,7 @@
 %global debug_package %{nil}
 
+# BTRFS is enabled by default, but can be disabled by defining _without_btrfs
+%{!?_with_btrfs: %{!?_without_btrfs: %define _with_btrfs 1}}
 
 Name: docker-ce
 Version: %{_version}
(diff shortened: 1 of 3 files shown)


thaJeztah:
ping @arkodg @tiborvass @silvin-lubecki PTAL

rpm/SPECS/docker-ce.spec
```diff
@@ -79,7 +81,7 @@ export DOCKER_GITCOMMIT=%{_gitcommit}
 mkdir -p /go/src/github.com/docker
 ln -s /root/rpmbuild/BUILD/src/engine /go/src/github.com/docker/docker
 
-pushd engine
+pushd /root/rpmbuild/BUILD/src/engine
```
silvin-lubecki: nit: can't we use an RPM variable here, like `$RPM_BUILD_ROOT` ?

thaJeztah: I'll look at some cleaning up in a follow-up 👍

mattvw:
Just curious: How long do you expect it will take to start seeing the CentOS 8 RPMs in the docker-de repo (e.g. under https://download.docker.com/linux/centos/)? Will we need to wait until the next docker-ce release? Thanks! :)

thaJeztah:
Yes, the plan is to add CentOS 8 for the next release, not in a 19.03.x patch release

PR #445, opened 2020-03-23
rpm: various fixes and clean-ups

thaJeztah:
### rpm: replace `%{buildroot}` with `$RPM_BUILD_ROOT`

From the [RPM mailinglist][1]:

> `$RPM_BUILD_ROOT` is the official, supported, mechanism for getting the
> value of the configured build root in a build scriptlet.
>
> `%{buildroot}` may be changed in the future w/o warning. Not that I'm planning
> on that, but you've been warned.

### rpm: fix double slashes in paths

The built-in macros for paths all have a leading slash, so removing slashes that were manually added before them.

### rpm: use macros/env-vars instead of hard-coded paths

Use the [built-in macros][2] and env-vars for some paths:

- Use `%{_libexecdir}` macro, instead of `/usr/libexec`
- Use `%{_datadir}` instead of `/usr/share`
- Use `%{_specsdir}` instead of `/root/rpmbuild/SPECS`
- Use `$RPM_BUILD_DIR` instead of `/root/rpmbuild/BUILD`
- Use `$RPM_SOURCE_DIR` instead of `/root/rpmbuild/SOURCES`

### rpm: consistently use curly brackets for env-vars

[1]: https://www.redhat.com/archives/rpm-list/2002-July/msg00121.html
[2]: https://rpm.org/user_doc/macros.html

### DIFF
--- a/rpm/SPECS/docker-ce-cli.spec
+++ b/rpm/SPECS/docker-ce-cli.spec
@@ -62,13 +62,13 @@ popd
 
 %install
 # install binary
-install -d $RPM_BUILD_ROOT%{_bindir}
-install -p -m 755 cli/build/docker $RPM_BUILD_ROOT%{_bindir}/docker
+install -d ${RPM_BUILD_ROOT}%{_bindir}
+install -p -m 755 cli/build/docker ${RPM_BUILD_ROOT}%{_bindir}/docker
 
 # install plugins
 pushd ${RPM_BUILD_DIR}/src/plugins
 for installer in *.installer; do
-    DESTDIR=$RPM_BUILD_ROOT \
+    DESTDIR=${RPM_BUILD_ROOT} \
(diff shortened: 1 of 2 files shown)


thaJeztah:
@silvin-lubecki PTAL

rpm/SPECS/docker-ce-cli.spec
```diff
@@ -62,33 +62,33 @@ popd
 
 %install
 # install binary
-install -d $RPM_BUILD_ROOT/%{_bindir}
```
silvin-lubecki: @thaJeztah isn't `%{_binddir}` already pointing to the right directory and self-sufficient?

silvin-lubecki: Oups 😅

PR #447, opened 2020-04-01
rpm: remove devicemapper requirement

thaJeztah:
devicemapper is now an optional configuration, so we shouldn't make it a required dependency.


thaJeztah:
ping @tiborvass @silvin-lubecki PTAL

PR #449, opened 2020-04-02
Replace the default common commit by a specific one for cli and engin…

glours:
…e for RPM  and Debian packaging

Signed-off-by: Guillaume Lours <guillaume.lours@docker.com>

### DIFF
--- a/deb/Makefile
+++ b/deb/Makefile
@@ -4,6 +4,8 @@ CLI_DIR=$(realpath $(CURDIR)/../../cli)
 ENGINE_DIR=$(realpath $(CURDIR)/../../engine)
 PLUGINS_DIR=$(realpath $(CURDIR)/../plugins)
 GITCOMMIT?=$(shell cd $(CLI_DIR) && git rev-parse --short HEAD)
+CLI_GITCOMMIT?=$(GITCOMMIT)
+ENGINE_GITCOMMIT?=$(GITCOMMIT)
 GO_BASE_IMAGE=golang
 GO_IMAGE?=$(GO_BASE_IMAGE):$(GO_VERSION)-buster
 CHOWN:=docker run --rm -v $(CURDIR):/v -w /v alpine chown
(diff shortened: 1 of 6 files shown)


deb/Makefile
```diff
@@ -3,7 +3,8 @@ include ../common.mk
 CLI_DIR=$(realpath $(CURDIR)/../../cli)
 ENGINE_DIR=$(realpath $(CURDIR)/../../engine)
 PLUGINS_DIR=$(realpath $(CURDIR)/../plugins)
-GITCOMMIT?=$(shell cd $(CLI_DIR) && git rev-parse --short HEAD)
+CLI_GITCOMMIT?=$(shell cd $(CLI_DIR) && git rev-parse --short HEAD)
+ENGINE_GITCOMMIT?=$(shell cd $(ENGINE_DIR) && git rev-parse --short HEAD)
```
thaJeztah, line 7: What we should do (to be sure we don't break anything), is

- if `CLI_GITCOMMIT` is not set, but `GITCOMMIT` is, then `CLI_GITCOMMIT=GITCOMMIT`
- same for `ENGINE_GITCOMMIT`

thaJeztah:
It's failing here;

```
# Build the daemon and dependencies
DOCKER_GITCOMMIT=238887ef06 cd engine && PRODUCT=docker ./hack/make.sh dynbinary

error: .git directory missing and DOCKER_GITCOMMIT not specified
  Please either build with the .git directory accessible, or specify the
  exact (--short) commit hash you are building using DOCKER_GITCOMMIT for
  future accountability in diagnosing build issues.  Thanks!
```

Looks like this should be;

```
cd engine && DOCKER_GITCOMMIT=238887ef06 PRODUCT=docker ./hack/make.sh dynbinary
```

deb/common/rules
```diff
@@ -9,12 +9,12 @@ override_dh_gencontrol:
 
 override_dh_auto_build:
 	# Build the daemon and dependencies
-	cd engine && PRODUCT=docker ./hack/make.sh dynbinary
+	DOCKER_GITCOMMIT=$(ENGINE_GITCOMMIT) cd engine && PRODUCT=docker ./hack/make.sh dynbinary
```
thaJeztah, line 12:
```suggestion
	cd engine && DOCKER_GITCOMMIT=$(ENGINE_GITCOMMIT) PRODUCT=docker ./hack/make.sh dynbinary
```

deb/Makefile
glours: @thaJeztah are you sure for the `ENGINE_COMMIT`, shouldn't we do the `git rev-parse` from the engine directory to be sure we use the correct commit and not the default `GITCOMMIT` hash which can come from the `CLI_DIR`?

glours: Updated the `ENGINE_GITCOMMIT` too

PR #457, opened 2020-04-23
[master] fetch source from upstream docker/docker repository

thaJeztah:
Various changes in this PR:

- remove the VERSION from the Jenkinsfile, as it was just to set a default (which is already defined in the Makefile)
- use upstream https://github.com/docker/docker (https://github.com/moby/moby) repository to fetch the source, and add Makefile variables to allow overriding the repository and reference (branch/tag/commit) to build from
- Add `DOCKER_CLI_REPO`, `DOCKER_ENGINE_REPO`, `REF`, `DOCKER_CLI_REF` and `DOCKER_ENGINE_REF` make variables to allow overriding the repository and  reference to build from.
- if an `ENGINE_DIR` or `CLI_DIR` is specified, copy the source instead of cloning (to keep the source in a fixed location)

PR #461, opened 2020-04-30
[master] Jenkinsfile: use docker/docker instead of docker/engine

thaJeztah:
extracted from https://github.com/docker/docker-ce-packaging/pull/457

The release branches have moved back to the upstream docker/docker
repository, so we should use that as source instead of the docker/engine
fork (and rely on it being synced with upstream)


thaJeztah:
@tiborvass PTAL

thaJeztah:
Ah, dang; looks like I didn't split the commits correctly

thaJeztah:
Should be fixed now

PR #433, opened 2020-02-24
Backport master latest commits to 19.03

glours:
Backporting commits from `master` to be close as possible with the nightly docker-ce packaging in the `19.03` branch

backport of:

- https://github.com/docker/docker-ce-packaging/pull/357 Assorted improvements to Makefiles
- https://github.com/docker/docker-ce-packaging/pull/374 Dockerfiles: remove unused GO_VERSION
- https://github.com/docker/docker-ce-packaging/pull/390 plugins: remove (now) redundant .gitkeep
- https://github.com/docker/docker-ce-packaging/pull/395 Update packages list
- https://github.com/docker/docker-ce-packaging/pull/396 [TAR-1174][master] Turn Go11Module=off
- https://github.com/docker/docker-ce-packaging/pull/403 Dockerfiles: re-use DISTRO, SUITE build-args where possible
- https://github.com/docker/docker-ce-packaging/pull/404 deb: Only recommend aufs-tools on amd64
- https://github.com/docker/docker-ce-packaging/pull/411 Remove image-based build (for docker engine activate)
- https://github.com/docker/docker-ce-packaging/pull/420 Remove some remnants related to image-based builds
- https://github.com/docker/docker-ce-packaging/pull/421 Makefile: enable buildkit when building images
- https://github.com/docker/docker-ce-packaging/pull/426 Remove more remnants from image-based engine
- https://github.com/docker/docker-ce-packaging/pull/427 update list of reviewers in unir config
- https://github.com/docker/docker-ce-packaging/pull/462 Makefile: chown engine dir during cleanup

[TAR-1174]: https://docker.atlassian.net/browse/TAR-1174

PR #462, opened 2020-05-01
Makefile: chown engine dir during cleanup

thaJeztah:
to address https://github.com/docker/docker-ce-packaging/pull/433#issuecomment-613427895

we should have a look why this target is in the main makefile, and why we don't use `make -C static clean` (e.g.)


thaJeztah:
ping @tiborvass @silvin-lubecki PTAL

PR #468, opened 2020-05-01
[master] deb: use DEBIAN_FRONTEND=noninteractive for all dockerfiles

thaJeztah:
This was needed for Ubuntu 20.04, but doesn't hurt to use for other versions as well.


thaJeztah:
@silvin-lubecki @tiborvass PTAL

PR #470, opened 2020-05-08
[master] Build CentOS 8 on PR builds and fix GPG errors

thaJeztah:
see individual commits for details

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -20,7 +20,7 @@ test_steps = [
 		}
 	},
 	'rpm': { ->
-		stage('Centos 7 RPM Package') {
+		stage('Centos 7 and 8 RPM Packages') {
 			wrappedNode(label: 'ubuntu && x86_64', cleanWorkspace: true) {
 				try {
 					checkout scm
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -20,15 +20,15 @@ test_steps = [
...
+		stage('Centos 7 and 8 RPM Packages') {
 			wrappedNode(label: 'ubuntu && x86_64', cleanWorkspace: true) {
 				try {
 					checkout scm
 					sh('git clone https://github.com/docker/cli.git')
 					sh("git -C cli checkout $branch")
 					sh('git clone https://github.com/docker/docker.git engine')
 					sh("git -C engine checkout $branch")
-					sh('make -C rpm VERSION=0.0.1-dev ENGINE_DIR=$(pwd)/engine CLI_DIR=$(pwd)/cli centos-7')
+					sh('make -C rpm VERSION=0.0.1-dev ENGINE_DIR=$(pwd)/engine CLI_DIR=$(pwd)/cli centos-7 centos-8')
```
tiborvass, line 31: why not too parallel stages?

thaJeztah: Yes we might do that,  it already didn't do so for the deb stage, so it wouldn't save on the whole run (also provisioning the machine takes some time, and now we can reuse some build cache)

PR #471, opened 2020-05-21
docker-rootless-extras: add dockerd-rootless-setuptool.sh

AkihiroSuda:
dockerd-rootless-setuptool.sh was added to Moby in https://github.com/moby/moby/pull/40950


AkihiroSuda:
@thaJeztah @tiborvass @silvin-lubecki PTAL?

PR #476, opened 2020-07-08
Remove Debian 9 "Stretch" (EOL: 2020-07-06)

AkihiroSuda:
Debian GNU/Linux 9 "Stretch" reached EOL on 2020-07-06: https://wiki.debian.org/DebianReleases


thaJeztah:
Seems good to me; want to double-check if we should take the LTS into account (https://wiki.debian.org/LTS, which seems to be July 6, 2020 to June 30, 2022)


@tianon any suggestions?

tianon:
I think this makes sense for the 20.xx release, but I don't think it should be backported to the 19.xx series.

For some context, LTS in Debian does not mean the same as LTS in Ubuntu.  In Debian, Stretch EOL means the Debian Security Team no longer maintains security fixes in Debian.  That is now the responsibility/purview of the "LTS" team, which is a group of volunteers who are paid by donations to Debian LTS (which is controversial within the Debian community in itself).  This is in contrast to Ubuntu where "LTS" means the same thing as supported, and it's just a longer timeframe than the intermediate releases.

To put this in a different context (which I think might be useful/helpful), the official images would officially recommend against *new* versions including a `stretch` variant at this point, or basing images off `stretch`, but will still typically maintain `stretch` variants for old supported versions.  As a concrete example, `php:7.3-apache-stretch` will probably be supported for a while, but `php:8.0-apache-stretch` will not exist.

So TL;DR, I think any 19.03.13 release should continue to support stretch, but 20.xx should not.  I guess what that likely means for this PR is that it LGTM, but I don't think it should be cherry-picked to the 19.03 branch? (Assuming this repo is used/maintained the same way as most other Docker repos. :sweat_smile:)

thaJeztah:
Thanks @tianon ! Yes, I think that makes sense. For other distros reaching EOL, we would normally remove builds for them (as people would otherwise run into issues anyway with packages no longer being there), but if this Debian version continues "functional" (at least for the coming months), I'm good with keeping it for 19.03 (so not cherry-pick)

thaJeztah:
I'll take @tianon's LGTM and merge this one 😅

PR #480, opened 2020-07-15
[master] Update minimum containerd.io version to v1.3.0

thaJeztah:
Relates to https://github.com/moby/moby/pull/41210

The engine now defaults to the "io.containerd.runc.v2" shim, which is only available in containerd v1.3.0 and up.

Opening as draft, because no containerd.io v1.3.x packages are available yet, so verification will likely fail

### DIFF
--- a/deb/common/control
+++ b/deb/common/control
@@ -27,7 +27,7 @@ Vcs-Git: git://github.com/docker/docker.git
 
 Package: docker-ce
 Architecture: linux-any
-Depends: containerd.io (>= 1.2.2-3),
+Depends: containerd.io (>= 1.3.0),
          docker-ce-cli,
          iptables,
          libseccomp2 (>= 2.3.0),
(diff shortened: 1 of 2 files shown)


deb/common/control
```diff
@@ -27,7 +27,7 @@ Vcs-Git: git://github.com/docker/docker.git
 
 Package: docker-ce
 Architecture: linux-any
-Depends: containerd.io (>= 1.2.2-3),
+Depends: containerd.io (>= 1.3.0),
```
thaJeztah, line 30: Need to check if these expect the `-x` suffix once we have packages for containerd v1.3.x available

tonistiigi (approved):

tiborvass (approved):

deb/common/control
thaJeztah: Discussing in our maintainers meeting; and should probably be fine without

PR #490, opened 2020-08-04
plugins: don't require "experimental" for buildx

thaJeztah:
Some features of buildx may still be "experimental" / non-final, but we can call those out individually if needed.

This change removes the requirement to enable "experimental" mode in the cli configuration file, so that it's easier for users to try buildx.

### DIFF
--- a/plugins/buildx.installer
+++ b/plugins/buildx.installer
@@ -18,8 +18,7 @@ build() {
         git fetch --all
         git checkout -q "${BUILDX_COMMIT}"
         local LDFLAGS
-        # TODO: unmark `-tp` when no longer a technical preview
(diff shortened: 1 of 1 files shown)


thaJeztah:
@tonistiigi @tiborvass @chris-crone @silvin-lubecki PTAL

hakman:
Somewhat related, any thoughts about `docker manifest`? Will it be removed or maybe become non-experimental too?

thaJeztah:
Still being discussed; we may remove the "experimental" requirement itself, but there's still (UX) issues with those subcommands that should be looked into, so they may still be marked "experimental" (in the documentation), to indicate that backward-compatibility of the commands is not guaranteed.

thaJeztah:
> We should update the buildx README too

@chris-crone yes, I had that one open in a tab; should be updated that starting with 20.x it's not needed to set the configuration (unless we decide to backport this change to 19.03, in which case the patch release version should be mentioned)

PR #491, opened 2020-08-08
[master] deb: make docker-ce "Provide" docker.io

thaJeztah:
fixes https://github.com/docker/for-linux/issues/1077

To satisfy the dependency for packages that "Depends: docker.io"

We need to verify if this causes problems for the docker-ce-cli package,
because that package has a Conflict with docker.io, which could mean that
adding the "Provides" would actually cause the docker-ce package to conflict
with the docker-ce-cli package;

Referring to the Debian packaging documentation:
https://www.debian.org/doc/debian-policy/ch-relationships.html#virtual-packages-provides

> ## 7.5. Virtual packages - Provides
>
> ...
> If there are both concrete and virtual packages of the same name, then the
> dependency may be satisfied (**or the conflict caused**) by either the
> concrete package with the name in question or any other concrete package
> which provides the virtual package with the name in question.

### DIFF
--- a/deb/common/control
+++ b/deb/common/control
@@ -46,6 +46,7 @@ Conflicts: docker (<< 1.5~),
            docker.io,
            lxc-docker,
            lxc-docker-virtual-package
+Provides: docker.io
 Replaces: docker-engine
 Description: Docker: the open-source application container engine
  Docker is a product for you to build, ship and run any application as a


thaJeztah:
Don't merge yet (see the description and my comment on https://github.com/docker/for-linux/issues/1077#issuecomment-670856489). Opening to have this available for review, but we can close if this doesn't work

SuperSandro2000:
This would work if only one package replaces, conflicts and provides a package but since docker-ce-cli conflicts with docker.io this wouldn't allow both docker-ce and docker-ce-cli to be installed.

A simple fix would be that docker-ce-cli wouldn't provide docker.io but requires the same version of docker-ce.

tianon:
This seems strange to me -- it feels wrong for another package to `Depends: docker.io`.  At most, I would expect `Recommends:` in a relation against Docker CLI / Engine.

seemethere:
> 
> 
> This would work if only one package replaces, conflicts and provides a package but since docker-ce-cli conflicts with docker.io this wouldn't allow both docker-ce and docker-ce-cli to be installed.
> 
> A simple fix would be that docker-ce-cli wouldn't provide docker.io but requires the same version of docker-ce.

It's also a bit backwards too since `docker.io` is technically both the daemon and the cli meaning that if someone gets into a state where they've only installed `docker-ce` it could leave them in a broken state anyways.

Which package is he referencing that has `Depends: docker.io`?

thaJeztah:
Let me close this one, per the discussion above

PR #494, opened 2020-08-18
fix rpmbuilding failed with btrfs disable

lining2020:
Fixes: https://github.com/docker/for-linux/issues/1087

Signed-off-by: Li Ning <lining_yewu@cmss.chinamobile.com>


rpm/SPECS/docker-ce.spec
```diff
@@ -84,6 +84,10 @@ export DOCKER_GITCOMMIT=%{_gitcommit_engine}
 mkdir -p /go/src/github.com/docker
 ln -s ${RPM_BUILD_DIR}/src/engine /go/src/github.com/docker/docker
 
+%if 1%{?_with_btrfs}
+export DOCKER_BUILDTAGS='exclude_graphdriver_btrfs'
```
thaJeztah, line 88: Looks like this will overwrite the existing `DOCKER_BUILDTAGS` ? https://github.com/docker/docker-ce-packaging/blob/1c4fdf511936510a893a9c1c092191bed3e0fb95/rpm/centos-8/Dockerfile#L14

lining2020: Hi @thaJeztah thanks for reviewing my PR.  I think we can append the DOCKER_BUILDTAGS.
I will give a patch v2.

lining2020:
@thaJeztah Should I also remove the duplicate env setting `exclude_graphdriver_btrfs ` in Dockerfile in my PR, and what do you think?  : )
https://github.com/docker/docker-ce-packaging/blob/master/rpm/centos-8/Dockerfile#L14

lining2020:
Many thanks if anyone can help the PR land.  : )

rpm/SPECS/docker-ce.spec
```diff
@@ -84,6 +84,11 @@ export DOCKER_GITCOMMIT=%{_gitcommit_engine}
 mkdir -p /go/src/github.com/docker
 ln -s ${RPM_BUILD_DIR}/src/engine /go/src/github.com/docker/docker
 
+%if 1%{?_with_btrfs}
+DOCKER_BUILDTAGS+=' exclude_graphdriver_btrfs'
+export DOCKER_BUILDTAGS
+%endif
```
thaJeztah, line 90: Wouldn't this already be set though? https://github.com/docker/docker-ce-packaging/blob/27f81645524d59f5d3cce2b433125c19589c5b17/rpm/centos-8/Dockerfile#L14

Also, shouldn't this be using the `_without_btrfs` option? https://github.com/docker/docker-ce-packaging/blob/4cd3a7eca2f7887b15e98c2e67be1880f337d291/rpm/SPECS/docker-ce.spec#L3

PR #496, opened 2020-08-27
[master] deb: update apt cache before running mk-build-deps

thaJeztah:
mk-build-deps does not seem to be updating the cache by itself, and
because of that may be installing packages using an outdated package index
if the previous layer is in the build-cache:

    #6 [stage-1 2/8] RUN apt-get update && apt-get install -y curl devscripts e...
    #6 CACHED

    #8 [stage-1 3/8] COPY common /root/build-deb/debian
    #8 CACHED

    #9 [stage-1 4/8] RUN mk-build-deps -t "apt-get -o Debug::pkgProblemResolver...


thaJeztah:
@tiborvass @StefanScherer PTAL

StefanScherer (approved):
LGTM 🆕

PR #501, opened 2020-10-05
systemd: set OOMScoreAdjust for dockerd

thaJeztah:
dockerd currently sets the oom-score-adjust itself. This functionality was added when we did not yet run dockerd as a systemd service. Now that we do, it's better to instead have systemd handle this.


silvin-lubecki (approved):
LGTM

PR #465, opened 2020-05-01
static: add containerd-shim-runc-v2

AkihiroSuda:
containerd-shim-runc-v2 is used in cgroup v2 mode

PR #504, opened 2020-10-09
[19.03 backport] static: add containerd-shim-runc-v2

hakman:
Follow-up to the discussion in https://github.com/docker/docker-ce-packaging/pull/465#issuecomment-706133455.

Signed-off-by: Akihiro Suda <akihiro.suda.cz@hco.ntt.co.jp>
(cherry picked from commit 912015018bf457c02afd87723fb8a7e42fee9a2f)

### DIFF
--- a/static/Makefile
+++ b/static/Makefile
@@ -25,7 +25,7 @@ static: static-linux cross-mac cross-win cross-arm ## create all static packages
 static-linux: static-cli static-engine ## create tgz with linux x86_64 client and server
 	mkdir -p build/linux/docker
 	cp $(CLI_DIR)/build/docker build/linux/docker/
-	for f in dockerd containerd ctr containerd-shim docker-init docker-proxy runc; do \
+	for f in dockerd containerd ctr containerd-shim containerd-shim-runc-v2 docker-init docker-proxy runc; do \
 		cp -L $(ENGINE_DIR)/bundles/binary-daemon/$$f build/linux/docker/$$f; \
 	done
 	tar -C build/linux -c -z -f build/linux/docker-$(STATIC_VERSION).tgz docker


thaJeztah (approved):
LGTM

@AkihiroSuda @silvin-lubecki ptal

AkihiroSuda (approved):

silvin-lubecki (approved):
LGTM

hakman:
Thanks everyone! :)

hakman:
@thaJeztah any plans for one more 19.03 release before switching to 20.10?

thaJeztah:
Yes, there likely will be a patch release for 19.03 (there's some  back ports in queue; https://github.com/moby/moby/pulls?q=is%3Apr+milestone%3A19.03.14). No date was set for that yet.

We also likely will have some overlap between 20.10 reaching GA, and 19.03 becoming EOL to allow for a smoother transition for users to upgrade; backporting will be more conservative during the overlap (limiting to "critical" and "security" fixes only)

hakman:
Thanks a lot for the explanation. 19.03 is quite stable these days and is nice to see it supported for so long.

hakman:
For some reason, I still can't find the `containerd-shim-runc-v2` binary in the latest static package https://download.docker.com/linux/static/stable/x86_64/docker-19.03.14.tgz. Any thoughts on what happened?

thaJeztah:
Hmmm... I see the problem;

```
make[2]: Leaving directory '/home/ubuntu/workspace/docker-ce-packaging_PR-504/engine'
mkdir -p build/linux/docker
cp /home/ubuntu/workspace/docker-ce-packaging_PR-504/cli/build/docker build/linux/docker/
for f in dockerd containerd ctr containerd-shim containerd-shim-runc-v2 docker-init docker-proxy runc; do \
	cp -L /home/ubuntu/workspace/docker-ce-packaging_PR-504/engine/bundles/binary-daemon/$f build/linux/docker/$f; \
done
cp: cannot stat '/home/ubuntu/workspace/docker-ce-packaging_PR-504/engine/bundles/binary-daemon/containerd-shim-runc-v2': No such file or directory
tar -C build/linux -c -z -f build/linux/docker-0.0.0-20201006212710-233a6379e5.tgz docker
# extra binaries for running rootless
mkdir -p build/linux/docker-rootless-extras
for f in rootlesskit rootlesskit-docker-proxy dockerd-rootless.sh vpnkit; do \
	if [ -f /home/ubuntu/workspace/docker-ce-packaging_PR-504/engine/bundles/binary-daemon/$f ]; then \
		cp -L /home/ubuntu/workspace/docker-ce-packaging_PR-504/engine/bundles/binary-daemon/$f build/linux/docker-rootless-extras/$f; \
	fi \
done
```

Interestingly, make did not _fail_ on that; looks like it may need something similar to https://github.com/moby/sys/pull/19#discussion_r445698069


Looking at the "install" script used in the Dockerfile for 19.03; https://github.com/moby/moby/blob/v19.03.14/hack/dockerfile/install/containerd.installer#L33-L35

It does not have this patch; https://github.com/moby/moby/pull/40658

hakman:
Looks like I have to do a bit of testing and come back with another PR. Thanks @thaJeztah.

thaJeztah:
Let me open a PR to add `.POSIX` to the makefile

thaJeztah:
hmm.. just the `.POSIX` didn't help catch the failing copy; looking further, perhaps the `.SHELLFLAGS = ...` approach might still be needed.

thaJeztah:
Arf.. nevermind. It's because I was running on macOS, and GNU make on macOS is really old (3.81), and doesn't support these options 😞

hakman:
Ahhh, right... I am always amazed by how old some dev tools can be on macOS.

thaJeztah:
I think it may have the same reason as it shipping an ancient version of Bash out of the box, which had to do with versions after that changing to GPLv3+

thaJeztah:
Anyway.. PR is up; https://github.com/docker/docker-ce-packaging/pull/510 😅

hakman:
Hehe, thanks! Was just looking at it. So, that should work on master, but when cherry-picked to 19.03, it should fail until the rest of the pieces are added, right?

thaJeztah:
Correct, yes; so doesn't fix the binary not being there, but at least makes the build fail in that case, so that it doesn't go unnoticed. Docker 19.03 itself does not depend on that binary, which is also why it didn't show up in other places.

In case it helps; if you're only looking for containerd, you can use the static packages for 20.10; those contain containerd v1.4.1 (not yet the latest 1.4.3 release); they're in the "test" channel as docker 20.10 is not yet GA, but the containerd binaries are non-beta releases (just happen to be only used by docker 20.10 at this moment, thus not yet in the "stable" channel); https://download.docker.com/linux/static/test/x86_64/

hakman:
I kind of missed this comment @thaJeztah. Thanks a lot for the tip. Should really help with getting kOps released with some good ARM64 support for both Docker and containerd.

Not sure about this change though. As 19.03 will still be around, maybe it should either be reverted or the other dependencies added.

PR #509, opened 2020-11-19
Add "procps" to daemon "Recommends" (for "docker top")

tianon:
~~Hi!  This is adjusting the dependencies for `docker-ce` and `docker-ce-cli` -- moving `git` to `Recommends` on the latter (because these days, `dockerd` should only be shelling out to `git` as a not-recommended-for-use backwards compatibility measure) and~~ adding `procps` (which is needed for `docker top` and given it's only `Priority: important` not essential, might not be installed).

Both of these *do* belong in `Recommends:` (not `Depends:`) because they're both related to technically optional behavior of the CLI (although behavior that's common enough to warrant `Recommends:` instead of `Suggests:` so most users get them automatically).

I'm guessing there are more items in this list of `Recommends:` for `docker-ce` that should *also* move, but I'm not personally up-to-date enough to recognize them offhand (`libltdl7`?).  Happy to adjust/include them if someone else knows more reliably! :heart:

(Directly inspired by https://github.com/bottlerocket-os/bottlerocket/pull/1210; @samuelkarp :heart:)

### DIFF
--- a/deb/common/control
+++ b/deb/common/control
@@ -38,6 +38,7 @@ Recommends: apparmor,
             git,
             libltdl7,
             pigz,
+            procps,
             xz-utils
 Suggests: aufs-tools [amd64], cgroupfs-mount | cgroup-lite
 Conflicts: docker (<< 1.5~),


tianon:
(and for the avoidance of doubt, yes, I did run this file through `wrap-and-sort` to ensure the formatting was consistent with what it prefers before submitting :sweat_smile:)

samuelkarp (dismissed):

thaJeztah:
@tonistiigi ptal

tonistiigi:
Git is required for `docker build` to work. In legacy builder it is a requirement on both client and daemon side. In buildkit, only daemon side.

deb/common/control
```diff
@@ -61,6 +60,7 @@ Description: Docker: the open-source application container engine
 Package: docker-ce-cli
 Architecture: linux-any
 Depends: ${shlibs:Depends}
+Recommends: git, procps
```
thaJeztah, line 63: Shouldn't `procps` be in the `docker-ce` package above (not in the cli)?

tianon:
> Git is required for `docker build` to work. In legacy builder it is a requirement on both client and daemon side. In buildkit, only daemon side.

This is hugely surprising to me -- there are a lot of benefits to what was implemented in https://github.com/moby/moby/pull/4364, the primary one being https://github.com/moby/moby/issues/2715.

deb/common/control
tianon: Oh right -- I got confused because `docker top` is a CLI command, but it really runs `ps` from the daemon context; will update!

thaJeztah:
@tianon had a chat about that with Tonis, and with buildkit the ssh-agent can be forwarded to the daemon, so giving the best of both worlds (git clone directly at the daemon side, without having to clone client-side and then sending), and the credentials not leaving the client.

tonistiigi:
Building from git is very important, not only for performance but for properly tracking build sources for remote caching and future inspection, repeatability and replay features. Having no client-side dependencies is also important for the language extensibility and for supporting different environments where builds can be invoked. Building from git is the preferred way to invoke a build and we'll only make it more common over time. 

The implementation itself may change in the future, eg. if there is movement with moby/buildkit#1048 , but atm it relies on a git package dependency.

deb/common/control
```diff
@@ -35,9 +35,9 @@ Depends: containerd.io (>= 1.4.1),
 Recommends: apparmor,
             ca-certificates,
             docker-ce-rootless-extras,
-            git,
```
thaJeztah, line 38: Following the discussion, looks like this one should stay, correct?

tianon: Yeah, I guess. Invoking `git clone` daemon-side (potentially with root permissions, if we aren't super careful) still feels really, really wrong to me, but I'll update.

tianon:
Updated to only add `procps` (and updated PR title/description accordingly).

thaJeztah (approved):
LGTM

thaJeztah:
bringing this one in

PR #511, opened 2020-12-02
[19.03 backport] systemd unit fixes/updates

thaJeztah:
Backports of:

- (picked to get a clean cherry-pick) https://github.com/docker/docker-ce-packaging/pull/488 systemd: add multi-user.target to After list
    - relates to https://github.com/moby/moby/pull/41297
- https://github.com/docker/docker-ce-packaging/pull/508 Do not "Bind" docker "To" containerd (carry #365)
    - fixes to https://github.com/docker/for-linux/issues/678
    - fixes/addresses/relates to https://github.com/docker/for-linux/issues/1155
    - also see https://bugs.launchpad.net/ubuntu/+source/containerd/+bug/1870514

### DIFF
--- a/systemd/docker.service
+++ b/systemd/docker.service
@@ -2,7 +2,7 @@
 Description=Docker Application Container Engine
 Documentation=https://docs.docker.com
 BindsTo=containerd.service
-After=network-online.target firewalld.service containerd.service
+After=network-online.target firewalld.service containerd.service multi-user.target
 Wants=network-online.target
 Requires=docker.socket
 
--- a/systemd/docker.service
+++ b/systemd/docker.service
@@ -1,10 +1,9 @@
 [Unit]
 Description=Docker Application Container Engine
 Documentation=https://docs.docker.com
-BindsTo=containerd.service
 After=network-online.target firewalld.service containerd.service multi-user.target
 Wants=network-online.target
-Requires=docker.socket
+Requires=docker.socket containerd.service
 
 [Service]
 Type=notify


thaJeztah:
@tianon @chris-crone @tiborvass PTAL

FWIW; I can drop the first cherry-pick, and resolve the merge conflict if we don't want that change

tianon:
cc @mwhudson @mitechie - I think you (or someone in your sphere) was doing some testing around this to verify that it's sufficient?  Any update you can report?

thaJeztah:
from https://github.com/docker/for-linux/issues/1155#issuecomment-737234388 it looks like the original issue would only be hit my users running the distro -packages, but definitely interested if there's other recommendations/fixes we should have

tianon:
See https://code.launchpad.net/~bryce/ubuntu/+source/docker.io/+git/docker.io/+merge/394913 (@bryceharrington :wave:) -- it sounds like in Canonical's testing, `Wants=containerd.service` (instead of `Requires=`) has the more appropriate behavior?

tianon:
(Basically, allowing `containerd` to restart freely without causing Docker to stop/restart also, which AFAIR Docker is supposed to be tolerant of?)

sergiodj:
> See https://code.launchpad.net/~bryce/ubuntu/+source/docker.io/+git/docker.io/+merge/394913 (@bryceharrington wave) -- it sounds like in Canonical's testing, `Wants=containerd.service` (instead of `Requires=`) has the more appropriate behavior?

Yeah, this is what we found to work best.  We've also identified a problem in docker.io's package that makes it not honour the "don't restart the docker service when upgrading the package" setting, so we're fixing that as well.

tianon:
Right, that one's a Canonical/Ubuntu-specific one though -- the upstream packages restart the daemon (intentionally) because otherwise there are often issues connecting to/managing it, or it has issues starting new containers, etc. and the official stance is that containers that should be restarted should have an appropriate restart policy. :sweat_smile:

(Thanks for commenting/confirming; really appreciate it! :+1:)

sergiodj:
> > See https://code.launchpad.net/~bryce/ubuntu/+source/docker.io/+git/docker.io/+merge/394913 (@bryceharrington wave) -- it sounds like in Canonical's testing, `Wants=containerd.service` (instead of `Requires=`) has the more appropriate behavior?
> 
> Yeah, this is what we found to work best. We've also identified a problem in docker.io's package that makes it not honour the "don't restart the docker service when upgrading the package" setting, so we're fixing that as well.

To better clarify my answer, we noticed that even with `Requires=` the relationship between docker and containerd was too strong, and it was making docker stop when containerd was being upgraded.

sergiodj:
> Right, that one's a Canonical/Ubuntu-specific one though -- the upstream packages restart the daemon (intentionally) because otherwise there are often issues connecting to/managing it, or it has issues starting new containers, etc. and the official stance is 
that containers that should be restarted should have an appropriate restart policy. sweat_smile

Right, this is Ubuntu-specific :-).

> (Thanks for commenting/confirming; really appreciate it! +1)

No problem!

tianon:
Opened https://github.com/docker/docker-ce-packaging/pull/512 :+1:

thaJeztah:
> To better clarify my answer, we noticed that even with Requires= the relationship between docker and containerd was too strong, and it was making docker stop when containerd was being upgraded.

I was discussing that earlier this week, abs was reminded that with one of the containerd updates this was actually required (otherwise an old client was kept in memory)

thaJeztah:
oh, actually, that problem was with containerd https://github.com/docker/containerd-packaging/pull/113

pvizeli:
Please don't merge this. It's just not correct for systemd: https://github.com/moby/moby/issues/41772

thaJeztah:
Thanks! Yes, we're aware of that issue, and will be publishing updated packages for 20.10 with the change reverted. 

This PR is a backport for 19.03 (if we do another patch release, but there's no ETA yet for that), but the revert of that change will be included here before it would go out.

thaJeztah:
I'm pushing a PR to revert for master / 20.10 later

thaJeztah:
closing this, as there's no more releases planned for 19.03

PR #512, opened 2020-12-05
Move containerd.service from Requires= to Wants=

tianon:
Per the systemd.unit documentation:

> If this unit gets activated, the units listed will be activated as well. If one of the other units fails to activate, and an ordering dependency After= on the failing unit is set, this unit will not be started. Besides, with or without specifying After=, this unit will be stopped if one of the other units is explicitly stopped.
>
> Often, it is a better choice to use Wants= instead of Requires= in order to achieve a system that is more robust when dealing with failing services.

This should also be generally "safe" given we added `--containerd=/run/containerd/containerd.sock` to the flags we pass to `dockerd`.

See https://github.com/docker/docker-ce-packaging/pull/511#issuecomment-739088735.

### DIFF
--- a/systemd/docker.service
+++ b/systemd/docker.service
@@ -2,8 +2,8 @@
 Description=Docker Application Container Engine
 Documentation=https://docs.docker.com
 After=network-online.target firewalld.service containerd.service
-Wants=network-online.target
-Requires=docker.socket containerd.service
+Wants=network-online.target containerd.service
+Requires=docker.socket
 
 [Service]
 Type=notify


thaJeztah:
love the typo; its on theme 😇

tianon:
> love the typo; its on theme

I have *no* idea what you're talking about!! (it's fixed :innocent: :heart:)

tianon:
I have verified that with this change, I'm able to `systemctl restart containerd` without Docker going down (and that it does successfully reconnect to containerd).

thaJeztah:
Thanks! I actually recalled I left this as a comment for follow up on https://github.com/docker/docker-ce-packaging/pull/508

> We may want to look into using Wants= instead of Requires=, because
that allows docker to continue running if containerd is restarted, quoting
the systemd documentation:

>> Often, it is a better choice to use Wants= instead of Requires= in order
to achieve a system that is more robust when dealing with failing services.

> Given that docker will likely still fail if the containerd socket is not
present, startup will fail if containerd is not running, but if containerd
is restarted, the docker daemon may be able to try reconnecting.

thaJeztah (approved):
LGTM

thaJeztah:
@tianon looks like this needs a rebase 😅

thaJeztah:
Closing here, because the file was moved back to the moby repository. 
I asked @aiordache to carry this (move to moby), thanks!

PR #519, opened 2021-01-06
Circleci project setup

juraganhajat:


StefanScherer:
Thanks @juraganhajat, but we don't switch over to CircleCI.

PR #520, opened 2021-01-12
deb: make dh-systemd dependency optional as it's deprecated

thaJeztah:
dh-systemd has been integrated into debhelper, starting with version 9.20160709, and has been removed in Debian 11 "bullseye"

This patch updates the control file to not require it as a dependency on current versions of debian that ship with that version of debhelper

Related discussions:

[1] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=822670
[2] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=958585


thaJeztah:
@silvin-lubecki @tianon PTAL

deb/common/control
```diff
@@ -7,7 +7,7 @@ Build-Depends: bash,
                ca-certificates,
                cmake,
                dh-apparmor,
-               dh-systemd,
+               debhelper (>= 9.20160709) | dh-systemd,
```
tianon, line 10: Same comment as in https://github.com/docker/containerd-packaging/pull/212#discussion_r555897064 - I don't think we need to go to this level anymore. :smile:

```suggestion
               debhelper (>= 10~),
```

(and dropping `--with systemd` in `deb/common/rules`, which needs to be tested but should work)

thaJeztah:
@silvin-lubecki @chris-crone ptal

silvin-lubecki (approved):
LGTM

thaJeztah:
Let me bring this one in

thaJeztah:
> (and dropping --with systemd in deb/common/rules, which needs to be tested but should work)

Oh, forgot looking into that part; let me open up a follow-up so that we can test if it's no longer needed

PR #534, opened 2021-03-29
deb/rpm: add target for bundles archive as consumed by release-repo

tiborvass:
This fixes a discrepancy between the rpmbuild and debbuild directory structure
and allows for a simple bundles archive target ready to be uploaded and consumed
by release-repo.

release-packaging could call these instead directly instead.

Example:

```
$ cd rpm && make rpmbuild/bundles-ce-centos-8-amd64.tar.gz
```

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -1,4 +1,6 @@
 ARCH=$(shell uname -m)
+# if this fails on centos-based OS: yum install -y epel-release && yum install -y dpkg
+DPKG_ARCH=$(shell dpkg --print-architecture)
 BUILDTIME=$(shell date -u -d "@$${SOURCE_DATE_EPOCH:-$$(date +%s)}" --rfc-3339 ns 2> /dev/null | sed -e 's/ /T/')
 CHOWN:=docker run --rm -v $(CURDIR):/v -w /v alpine chown
 DEFAULT_PRODUCT_LICENSE:=Community Engine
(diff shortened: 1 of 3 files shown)


common.mk
```diff
@@ -1,4 +1,6 @@
 ARCH=$(shell uname -m)
+# if this fails on centos-based OS: yum install -y epel-release && yum install -y dpkg
+DPKG_ARCH=$(shell dpkg --print-architecture)
```
thaJeztah, line 3: Should this go in the deb/Makefile? (I think it's only used there, correct?)

tiborvass: it's already in that makefile and in rpm/Makefile as well via an include.

tiborvass: Ah sorry misunderstood. It is used in both deb and rpm Makefiles.

tiborvass: Looks like this is not super important, as nothing checks the architecture. This is what is consuming it: https://github.com/docker/release-repo/blob/031eeb8a6e4b99ab08b7c12e72f29a8d5dde64da/Makefile#L66

tiborvass: I just wanted to keep the same names as in release-packaging https://github.com/docker/release-packaging/blob/357e79197ac898fe1a2a89b3120c16df7de4a07b/Makefile#L108

thaJeztah (approved):
LGTM

PR #536, opened 2021-03-30
deb: create debbuild folder like it is done for rpm

tiborvass:
This is to fix a permissions issue with the release-packaging pipeline.

To fix issues in https://github.com/docker/release-packaging/pull/569.

Errors such as
```
tar (child): debbuild/bundles-ce-raspbian-buster-armhf.tar.gz: Cannot open: Permission denied
```

were because debbuild folder was created by docker run's `-v` giving it root ownership instead of current user.
This did not affect rpms because for some reason rpm/Makefile creates and chowns the folder correctly before calling docker run.

### DIFF
--- a/deb/Makefile
+++ b/deb/Makefile
@@ -68,9 +68,10 @@ raspbian: $(RASPBIAN_VERSIONS) ## build all raspbian deb packages
 .PHONY: $(DISTROS)
 $(DISTROS): sources/cli.tgz sources/engine.tgz sources/docker.service sources/docker.socket sources/plugin-installers.tgz
 	@echo "== Building packages for $@ =="
+	mkdir -p "debbuild/$@"
 	$(BUILD)
 	$(RUN)
-	$(CHOWN) -R $(shell id -u):$(shell id -g) debbuild/$@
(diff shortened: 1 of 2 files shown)


deb/Makefile
```diff
@@ -68,9 +68,11 @@ raspbian: $(RASPBIAN_VERSIONS) ## build all raspbian deb packages
 .PHONY: $(DISTROS)
 $(DISTROS): sources/cli.tgz sources/engine.tgz sources/docker.service sources/docker.socket sources/plugin-installers.tgz
 	@echo "== Building packages for $@ =="
+	mkdir -p "debbuild/$@"
+	$(CHOWN) -R root:root "debbuild/$@"
```
thaJeztah, line 72: Curious why it has to be owned by root here, if the container is running as root, it would have access, correct?

tiborvass: @thaJeztah I copied it from rpm/Makefile. It seems to work without it so I'll remove the root-chowning in both.

thaJeztah (approved):
LGTM

PR #540, opened 2021-04-05
mac: build arm64 binaries

tiborvass:

This PR uses golang 1.16 In order to build plugins for darwin/arm64.
However, docker app plugin does not build with go 1.16 (could not figure out how to disable go modules), and it was meant to be removed anyway.
ARM binaries had issues building with cgo, so as a workaround, this PR sets CGO_ENABLED=0 for arm (32 bit, aka armel/armhf) binaries only.

Related:
- https://github.com/docker/cli/pull/2993
- https://github.com/docker/release-packaging/pull/574
- https://github.com/docker/release-packaging/pull/568
- https://github.com/docker/release-repo/pull/575

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,5 +75,5 @@ deb: checkout ## build deb packages
 static: DOCKER_BUILD_PKGS:=static-linux cross-mac cross-win cross-arm
 static: checkout ## build static-compiled packages
 	for p in $(DOCKER_BUILD_PKGS); do \
-		$(MAKE) -C $@ VERSION=$(VERSION) GO_VERSION=$(GO_VERSION) $${p}; \
+		$(MAKE) -C $@ VERSION=$(VERSION) GO_VERSION=$(GO_VERSION) TARGETPLATFORM=$(TARGETPLATFORM) $${p}; \
 	done
--- a/deb/common/rules
+++ b/deb/common/rules
@@ -10,8 +10,7 @@ override_dh_auto_build:
 	cd engine && TMP_GOPATH="/go" hack/dockerfile/install/install.sh proxy dynamic
 	cd engine && TMP_GOPATH="/go" hack/dockerfile/install/install.sh rootlesskit dynamic
 	# Build the CLI
-	cd /go/src/github.com/docker/cli && \
-		LDFLAGS='' DISABLE_WARN_OUTSIDE_CONTAINER=1 make VERSION=$(VERSION) GITCOMMIT=$(CLI_GITCOMMIT) dynbinary manpages
+	cd /go/src/github.com/docker/cli && VERSION=$(VERSION) GITCOMMIT=$(CLI_GITCOMMIT) LDFLAGS='' GO_LINKMODE=dynamic ./scripts/build/binary && DISABLE_WARN_OUTSIDE_CONTAINER=1 LDFLAGS='' make manpages
 
 	# Build the scan-plugin
 	# TODO change once we support scan-plugin on other architectures
--- a/static/Makefile
+++ b/static/Makefile
@@ -74,7 +74,7 @@ cross-arm: cross-all-cli ## create tgz with linux armhf client only
 
 .PHONY: static-cli
 static-cli:
-	$(MAKE) -C $(CLI_DIR) -f docker.Makefile VERSION=$(GEN_STATIC_VER) build
+	cd $(CLI_DIR) && VERSION=$(GEN_STATIC_VER) docker buildx bake --set binary.platform=$(TARGETPLATFORM) --set binary.args.CGO_ENABLED=$(CGO_ENABLED) binary
 
 .PHONY: static-engine
 static-engine:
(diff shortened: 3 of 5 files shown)


deb/common/rules
```diff
@@ -10,8 +10,7 @@ override_dh_auto_build:
 	cd engine && TMP_GOPATH="/go" hack/dockerfile/install/install.sh proxy dynamic
 	cd engine && TMP_GOPATH="/go" hack/dockerfile/install/install.sh rootlesskit dynamic
 	# Build the CLI
-	cd /go/src/github.com/docker/cli && \
-		LDFLAGS='' DISABLE_WARN_OUTSIDE_CONTAINER=1 make VERSION=$(VERSION) GITCOMMIT=$(CLI_GITCOMMIT) dynbinary manpages
```
thaJeztah, line 14: Why did you remove the LDFLAGS='' here?

(perhaps we need the same comment as in the plugins install script?

tiborvass: I didn't think enough, and had too much confidence in Tonis's scripts :) I put it back.

deb/common/rules
```diff
@@ -10,8 +10,7 @@ override_dh_auto_build:
...
+	cd /go/src/github.com/docker/cli && VERSION=$(VERSION) GITCOMMIT=$(CLI_GITCOMMIT) LDFLAGS='' GO_LINKMODE=dynamic ./scripts/build/binary && DISABLE_WARN_OUTSIDE_CONTAINER=1 make manpages
```
thaJeztah, line 13:
```suggestion
	cd /go/src/github.com/docker/cli && VERSION=$(VERSION) GITCOMMIT=$(CLI_GITCOMMIT) LDFLAGS='' GO_LINKMODE=dynamic ./scripts/build/binary && LDFLAGS='' DISABLE_WARN_OUTSIDE_CONTAINER=1 make manpages
```

Need to set it for both `./scripts/build/binary` and for `make manpages` (we don't _export_ LDFLAGS, so it's not set after `&&` otherwise)

thaJeztah (changes requested):

static/Makefile
```diff
@@ -44,14 +43,23 @@ hash_files:
...
+buildx:
+	docker buildx inspect | grep -q 'Driver: docker-container' || docker buildx create --use
+
 .PHONY: cross-mac
-cross-mac: cross-all-cli cross-mac-plugins ## create tgz with darwin x86_64 client only
-	mkdir -p build/mac/docker
-	cp $(CLI_DIR)/build/docker-darwin-amd64 build/mac/docker/docker
-	tar -C build/mac -c -z -f build/mac/docker-$(GEN_STATIC_VER).tgz docker
+cross-mac: buildx cross-mac-plugins
+	cd $(CLI_DIR) && VERSION=$(GEN_STATIC_VER) docker buildx bake --set cross.platform=darwin/amd64,darwin/arm64 cross
```
tonistiigi, line 52: why are you building cross?

thaJeztah (approved):
LGTM after fixing the windows stage

Makefile
```diff
@@ -75,5 +75,5 @@ deb: checkout ## build deb packages
 static: DOCKER_BUILD_PKGS:=static-linux cross-mac cross-win cross-arm
 static: checkout ## build static-compiled packages
 	for p in $(DOCKER_BUILD_PKGS); do \
-		$(MAKE) -C $@ VERSION=$(VERSION) GO_VERSION=$(GO_VERSION) $${p}; \
+		$(MAKE) -C $@ VERSION=$(VERSION) GO_VERSION=$(GO_VERSION) GOPLATFORM=$(GOPLATFORM) $${p}; \
```
tonistiigi, line 78: why is this called `GOPLATFORM`?

static/Makefile
```diff
@@ -66,7 +73,7 @@ cross-arm: cross-all-cli ## create tgz with linux armhf client only
 
 .PHONY: static-cli
 static-cli:
-	$(MAKE) -C $(CLI_DIR) -f docker.Makefile VERSION=$(GEN_STATIC_VER) build
+	cd $(CLI_DIR) && VERSION=$(GEN_STATIC_VER) docker buildx bake --set binary.platform=$(GOPLATFORM) $(if $(findstring arm/,$(GOPLATFORM)),--set binary.args.CGO_ENABLED=0) binary
```
tonistiigi, line 76: Why don't you just pass in `CGO_ENABLED`? Afaics it is not that arm binaries can't use cgo but that arm call happens on some specific environment/jenkins node set from the caller side that causes it. If this would be arm specific we would disable cgo in `docker/cli` for that case(as we disable pie for some arch) but don't think but I don't think there is any proof of that atm.

Makefile
tiborvass: i can call it PLATFORM if you want but in this repo it's confusing as it can also mean the stupid "Docker CE" PLATFORM string. Happy to change the name.

thaJeztah: It's not GOARCH?

Could be GO_PLATFORM to be consistent with GO_VERSION

tiborvass:
@tonistiigi fixed your concerns.

tiborvass:
@tonistiigi arm workaround moved to https://github.com/docker/release-packaging/pull/574/files#diff-e6ffa5dc854b843b3ee3c3c28f8eae2f436c2df2b1ca299cca1fa5982e390cf8R226

tiborvass:
Ok this is green from https://ci-next.docker.com/teams-core/blue/organizations/jenkins/release-packaging/detail/PR-574/68/pipeline/

tonistiigi (approved):

thaJeztah (approved):
LGTM

PR #547, opened 2021-05-20
Use systemd unit files from upstream repository

thaJeztah:
The packaging repository was maintaining its own version of the systemd units.
These files were originally based on the units in the upstream docker / moby
repository, but diverged.

PR https://github.com/moby/moby/pull/42373 merged back the units from this
repository to upstream, so we can now consume those files instead of the
ones that were maintained in this repository.


fredericdalleau (approved):
LGTM

PR #548, opened 2021-05-25
Remove docker-app plugin

StefanScherer:
docker app is not maintained, and can sometimes cause performance problems.
We want to remove it from Docker Desktop.


thaJeztah:
Do you know if we wanted to add a "stub" script that would print instructions (or a link) on how to install `docker app` (if we want to include this in 20.10.7 patch release), or only mention in the changelog / release notes?

thaJeztah:
if we want to include this in 20.10.7, this needs to be cherry-picked into the 20.10 branch after merged to master

StefanScherer:
I don't think we need it in 20.10.7 patch release, no hurry. We handle that removal in the Docker Desktop build pipeline right now.

thaJeztah (approved):
LGTM

should at least be fine for master 👍

PR #552, opened 2021-06-02
buildx: remove "-docker" suffix from version

thaJeztah:
When building the CLI plugin, we were adding a "-docker" suffix. It's a bit
unclear if there was a specific reason for this (possibly to differentiate
between builds from GitHub and builds that are part of the CLI?

Initially, the version had a `-tp-docker` prefix (where `-tp` was used to
indicate it was still experimental); this prefix was removed in commit
0b45cce1abcf8868f4d957317a839f1e80058607 (https://github.com/docker/docker-ce-packaging/pull/490)

    /usr/libexec/docker/cli-plugins/docker-buildx docker-cli-plugin-metadata
    {
         "SchemaVersion": "0.1.0",
         "Vendor": "Docker Inc.",
         "Version": "v0.5.1-docker",
         "ShortDescription": "Build with BuildKit"
    }

This patch removes the `-docker` suffix from the version string.

### DIFF
--- a/plugins/buildx.installer
+++ b/plugins/buildx.installer
@@ -18,7 +18,7 @@ build() {
         git fetch --all
         git checkout -q "${BUILDX_COMMIT}"
         local LDFLAGS
-        LDFLAGS="-X ${PKG}/version.Version=$(git describe --match 'v[0-9]*' --always --tags)-docker -X ${PKG}/version.Revision=$(git rev-parse HEAD) -X ${PKG}/version.Package=${PKG}"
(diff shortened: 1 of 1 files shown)


thaJeztah:
@tonistiigi @chris-crone PTAL

chris-crone:
SGTM but maybe @tonistiigi has context that I don't

tonistiigi:
It was added to determine the releases that required experimental to be turned on. Github releases of Buildx never required it. Maybe we should keep this until the builds here do not build with the Dockerfile from the upstream repo and define their own build scripts so if there is an issue reported we can see what script made the binary.

thaJeztah:
Superseded by https://github.com/docker/docker-ce-packaging/pull/654, which also removed the suffix

PR #553, opened 2021-06-29
[master] add docker-compose-plugin package (deb, rpm)

thaJeztah:
depends on:

- [x] https://github.com/docker/docker-ce-packaging/pull/621

Adds a package for the docker-compose cli plugin

with this:

```bash
dpkg -i ./docker-ce-cli_0.0.0-20210629140245-4a6fe51-0~ubuntu-focal_amd64.deb
dpkg -i ./docker-compose-plugin_2.0.0-beta.6~ubuntu-focal_amd64.deb

docker info --format '{{json .ClientInfo.Plugins}}' | jq .
[
  {
    "SchemaVersion": "0.1.0",
    "Vendor": "Docker Inc.",
    "Version": "v0.5.1-docker",
    "ShortDescription": "Build with BuildKit",
    "Name": "buildx",
    "Path": "/usr/libexec/docker/cli-plugins/docker-buildx"
  },
  {
    "SchemaVersion": "0.1.0",
    "Vendor": "Docker Inc.",
    "Version": "v2.0.0-beta.6",
    "ShortDescription": "Docker Compose",
    "Name": "compose",
    "Path": "/usr/libexec/docker/cli-plugins/docker-compose"
  }
]
```

```bash
rpm -i \
    ./docker-ce-cli-0.0.0.20210702143511.f782d13-0.el8.x86_64.rpm \
    ./docker-scan-plugin-0.8.0-0.el8.x86_64.rpm \
    ./docker-compose-plugin-2.0.0.beta.6-0.el8.x86_64.rpm

docker info --format '{{json .ClientInfo.Plugins}}' | jq .
[
  {
    "SchemaVersion": "0.1.0",
    "Vendor": "Docker Inc.",
    "Version": "v0.5.1-docker",
    "ShortDescription": "Build with BuildKit",
    "Name": "buildx",
    "Path": "/usr/libexec/docker/cli-plugins/docker-buildx"
  },
  {
    "SchemaVersion": "0.1.0",
    "Vendor": "Docker Inc.",
    "Version": "v2.0.0-beta.6",
    "ShortDescription": "Docker Compose",
    "Name": "compose",
    "Path": "/usr/libexec/docker/cli-plugins/docker-compose"
  },
  {
    "SchemaVersion": "0.1.0",
    "Vendor": "Docker Inc.",
    "Version": "v0.8.0",
    "ShortDescription": "Docker Scan",
    "Name": "scan",
    "Path": "/usr/libexec/docker/cli-plugins/docker-scan"
  }
]
```

### DIFF
--- a/deb/common/control
+++ b/deb/common/control
@@ -102,6 +102,18 @@ Description: Rootless support for Docker.
   Either VPNKit or slirp4netns (>= 0.4.0) needs to be installed separately.
 Homepage: https://docs.docker.com/engine/security/rootless/
 
+Package: docker-compose-plugin
+Priority: optional
+Architecture: linux-any
+Enhances: docker-ce-cli
+Description: Docker Compose (V2) plugin for the Docker CLI.
+ .
+ This plugin provides the 'docker compose' subcommand.
+ .
+ The binary can also be run standalone as a direct replacement for
+ Docker Compose V1 ('docker-compose').
+Homepage: https://github.com/docker/compose
+
 Package: docker-scan-plugin
 Priority: optional
 # TODO change once we support scan-plugin on other architectures (see dpkg-architecture -L)
--- a/deb/common/rules
+++ b/deb/common/rules
@@ -16,6 +16,15 @@ override_dh_auto_build:
 	# Build the CLI
 	cd /go/src/github.com/docker/cli && VERSION=$(VERSION) GITCOMMIT=$(CLI_GITCOMMIT) LDFLAGS='' GO_LINKMODE=dynamic ./scripts/build/binary && DISABLE_WARN_OUTSIDE_CONTAINER=1 LDFLAGS='' make manpages
 
+	# Build the compose plugin
+	# FIXME: using GOPROXY, to work around:
+	# go: github.com/Azure/azure-sdk-for-go@v48.2.0+incompatible: reading github.com/Azure/azure-sdk-for-go/go.mod at revision v48.2.0: unknown revision v48.2.0
+	cd /go/src/github.com/docker/compose \
+	&& GOPROXY="https://proxy.golang.org" GO111MODULE=on go mod download \
+	&& mkdir -p /usr/libexec/docker/cli-plugins/ \
+	&& GOPROXY="https://proxy.golang.org" GO111MODULE=on GIT_TAG=$(COMPOSE_VERSION) \
+		make COMPOSE_BINARY=/usr/libexec/docker/cli-plugins/docker-compose -f builder.Makefile compose-plugin
+
 	# Build the scan-plugin
 	# TODO change once we support scan-plugin on other architectures
 	if [ "$(TARGET_ARCH)" = "amd64" ]; then \
@@ -39,6 +48,9 @@ override_dh_auto_test:
 
 	ver="$$(cli/build/docker --version)"; \
 		test "$$ver" = "Docker version $(VERSION), build $(CLI_GITCOMMIT)" && echo "PASS: cli version OK" || (echo "FAIL: cli version ($$ver) did not match" && exit 1)
+
+	ver="$$(/usr/libexec/docker/cli-plugins/docker-compose docker-cli-plugin-metadata | awk '{ gsub(/[",:]/,"")}; $$1 == "Version" { print $$2 }')"; \
+		test "$$ver" = "$(COMPOSE_VERSION)" && echo "PASS: docker-compose version OK" || (echo "FAIL: docker-compose version ($$ver) did not match" && exit 1)
 
 	# FIXME: --version currently doesn't work as it makes a connection to the daemon, so using the plugin metadata instead
 	# TODO change once we support scan-plugin on other architectures
@@ -69,6 +81,9 @@ override_dh_auto_install:
 	install -D -m 0755 $(shell readlink -e engine/bundles/dynbinary-daemon/docker-proxy) debian/docker-ce/usr/bin/docker-proxy
 	install -D -m 0755 /usr/local/bin/docker-init debian/docker-ce/usr/bin/docker-init
 
+	# docker-compose-plugin install
+	install -D -m 0755 /usr/libexec/docker/cli-plugins/docker-compose debian/docker-compose-plugin/usr/libexec/docker/cli-plugins/docker-compose
+
 	# docker-scan-plugin install
 	# TODO change once we support scan-plugin on other architectures
 	if [ "$(TARGET_ARCH)" = "amd64" ]; then \
@@ -95,6 +110,10 @@ override_dh_install:
 	dh_apparmor --profile-name=docker-ce -pdocker-ce
 
 override_dh_gencontrol:
+	# Use separate version for the compose-plugin package, then generate the other control files as usual
+	# TODO override "Source" field in control as well (to point to compose, as it doesn't match the package name)
+	dh_gencontrol -pdocker-compose-plugin -- -v$${COMPOSE_VERSION#v}~$${DISTRO}-$${SUITE}
+
 	# Use separate version for the scan-plugin package, then generate the other control files as usual
 	# TODO override "Source" field in control as well (to point to scan-cli-plugin, as it doesn't match the package name)
 	# TODO change once we support scan-plugin on other architectures (see dpkg-architecture -L)
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ clean-src:
 	$(RM) -r src
 
 .PHONY: src
-src: src/github.com/docker/cli src/github.com/docker/docker src/github.com/docker/scan-cli-plugin ## clone source
+src: src/github.com/docker/cli src/github.com/docker/docker src/github.com/docker/compose src/github.com/docker/scan-cli-plugin ## clone source
 
 ifdef CLI_DIR
 src/github.com/docker/cli:
@@ -37,6 +37,10 @@ src/github.com/docker/docker:
 	git -C $@ remote add origin "$(DOCKER_ENGINE_REPO)"
 endif
 
+src/github.com/docker/compose:
+	git init $@
+	git -C $@ remote add origin "$(DOCKER_COMPOSE_REPO)"
+
 src/github.com/docker/scan-cli-plugin:
 	git init $@
 	git -C $@ remote add origin "$(DOCKER_SCAN_REPO)"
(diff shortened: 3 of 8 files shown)


deb/common/rules
```diff
@@ -7,11 +7,19 @@ override_dh_auto_build:
...
-	cd engine && TMP_GOPATH="/go" hack/dockerfile/install/install.sh proxy dynamic
 	cd engine && TMP_GOPATH="/go" hack/dockerfile/install/install.sh rootlesskit dynamic
 	# Build the CLI
 	cd /go/src/github.com/docker/cli && VERSION=$(VERSION) GITCOMMIT=$(CLI_GITCOMMIT) LDFLAGS='' GO_LINKMODE=dynamic ./scripts/build/binary && DISABLE_WARN_OUTSIDE_CONTAINER=1 LDFLAGS='' make manpages
 
+	# Build the compose-cli plugin
+	# FIXME: using GOPROXY, to work around:
+	# go: github.com/Azure/azure-sdk-for-go@v48.2.0+incompatible: reading github.com/Azure/azure-sdk-for-go/go.mod at revision v48.2.0: unknown revision v48.2.0
+	cd /go/src/github.com/docker/compose-cli \
+	&& GOPROXY="https://proxy.golang.org" GO111MODULE=on go mod download \
```
thaJeztah, line 18: We really need to update the Dockerfiles to include the source-code in an (intermediate) stage, instead of bind-mounting from the host. Especially with go modules, we should cache those.

thaJeztah:
interesting failure;

```
shift
+ '[' '!' -f hack/dockerfile/install/tini.installer ']'
+ . hack/dockerfile/install/tini.installer
++ : de40ad007797e0dcd8b7126f27bb87401d224240
+ install_tini
+ echo 'Install tini version de40ad007797e0dcd8b7126f27bb87401d224240'
Install tini version de40ad007797e0dcd8b7126f27bb87401d224240
+ git clone https://github.com/krallin/tini.git /go/tini
Cloning into '/go/tini'...
+ cd /go/tini
+ git checkout -q de40ad007797e0dcd8b7126f27bb87401d224240
+ cmake .
cmake: symbol lookup error: cmake: undefined symbol: archive_write_add_filter_zstd
error: Bad exit status from /var/tmp/rpm-tmp.shIdtf (%build)
```

thaJeztah:
Rebased to include https://github.com/docker/docker-ce-packaging/pull/554 (to fix CI failures) only the last commit is currently new

thaJeztah:
Hmm.. I see `internal.Version` is set to `v2.0.0-beta.4`:

```bash
GIT_TAG=v2.0.0-beta.4
make COMPOSE_BINARY=bin/docker-compose -f builder.Makefile compose-plugin \
    GOOS=linux \
    GOARCH=amd64 \
    CGO_ENABLED=0 \
    go build \
        -trimpath \
        -ldflags="-s -w -X github.com/docker/compose-cli/internal.Version=v2.0.0-beta.4" \
        -o bin/docker-compose \
        ./cmd
```

However, the binary has the `v` prefix stripped:

```bash
/root/rpmbuild/BUILDROOT/docker-compose-plugin-2.0.0.beta.4-0.fc34.x86_64/usr/libexec/docker/cli-plugins/docker-compose docker-cli-plugin-metadata
++ awk '{ gsub(/[",:]/,"")}; $1 == "Version" { print $2 }'
+ ver=2.0.0-beta.4
+ test 2.0.0-beta.4 = v2.0.0-beta.4
FAIL: docker-compose version (2.0.0-beta.4) did not match
```

deb/common/rules
thaJeztah: Of course, this also shows I made a stupid mistake in the check (it doesn't exist with a non-zero exit code because the `echo` is "successful"); fixing that.

thaJeztah:
Of course, that also shows I made a stupid mistake in the check (it doesn't exist with a non-zero exit code because the `echo` is "successful"); fixing that.

thaJeztah:
Updated the check, and it works now:

```
++ /root/rpmbuild/BUILDROOT/docker-compose-plugin-2.0.0.beta.4-0.el7.x86_64/usr/libexec/docker/cli-plugins/docker-compose docker-cli-plugin-metadata
++ awk '{ gsub(/[",:]/,"")}; $1 == "Version" { print $2 }'
+ ver=2.0.0-beta.4
+ test 2.0.0-beta.4 = v2.0.0-beta.4
+ echo 'FAIL: docker-compose version (2.0.0-beta.4) did not match'
FAIL: docker-compose version (2.0.0-beta.4) did not match
+ exit 1

RPM build errors:
error: Bad exit status from /var/tmp/rpm-tmp.ywJz5g (%check)
    Bad exit status from /var/tmp/rpm-tmp.ywJz5g (%check)
make: *** [Makefile:97: centos-7] Error 1
make: Leaving directory '/home/ubuntu/workspace/docker-ce-packaging_PR-553/rpm'
```

```
ver="$(/usr/libexec/docker/cli-plugins/docker-compose docker-cli-plugin-metadata | awk '{ gsub(/[",:]/,"")}; $1 == "Version" { print $2 }')"; \
	test "$ver" = "v2.0.0-beta.4" && echo "PASS: docker-compose version OK" || (echo "FAIL: docker-compose version ($ver) did not match" && exit 1)
FAIL: docker-compose version (2.0.0-beta.4) did not match
debian/rules:41: recipe for target 'override_dh_auto_test' failed
make[1]: Leaving directory '/root/build-deb'
make[1]: *** [override_dh_auto_test] Error 1
make: *** [build] Error 2
debian/rules:118: recipe for target 'build' failed
dpkg-buildpackage: error: debian/rules build gave error exit status 2
make: *** [Makefile:77: ubuntu-xenial] Error 2
make: Leaving directory '/home/ubuntu/workspace/docker-ce-packaging_PR-553/deb'
script returned exit code 2
```

Will now push a commit to (temporarily) ignore the difference

thaJeztah:
> Will now push a commit to (temporarily) ignore the difference

Removed the temporary fix, because v2.0.0-beta.5 was tagged, which removes the `v` prefix https://github.com/docker/compose-cli/releases/tag/v2.0.0-beta.5

ndeloof:
build is broken as our release process created a tag on the master branch, not v2

ndeloof:
failure on `go install github.com/rootless-containers/rootlesskit/cmd/rootlesskit` ◔_◔

thaJeztah:
> failure on `go install github.com/rootless-containers/rootlesskit/cmd/rootlesskit` ◔_◔

Arf.. would that be because of https://github.com/moby/moby/commit/3cec4b8cd42179e67dcfdda2adce03516c6a5cd1 ? The issue linked to https://github.com/golang/go/issues/38373, but that mentions it _should be_ fixed in Go 1.14, but apparently it's not

> Also, old git clients have a bug for which the cmd/go workaround has probably not been backported. You may need to upgrade to Go 1.14.1 if you are affected.

Guess we need to add `GOPROXY="https://proxy.golang.org"` in the CentOS 7 Dockerfile

thaJeztah:
Opened https://github.com/docker/docker-ce-packaging/pull/577 for the rootlesskit failure

ndeloof:
@rfay we just published the "dash-to-space" docker compose redirection tool on https://github.com/docker/compose-switch
An install script will come soon (cc @ulyssessouza), but you already can give it a try - this tool assumes the legacy docker-compose is available on PATH as [docker-compose-v1](https://github.com/docker/compose-switch/blob/master/redirect/paths.go#L6)

rfay:
That's an odd requirement, as legacy docker-compose on Linux is never going to be available as docker-compose-v1... What's up with that? `docker-compose` on Linux is called `docker-compose`. In fact, AFAIK, there aren't any linux machines using the official Docker packages that implement `docker compose` either?

When you get this sorted out, is the tool going to be available in deb packages in your debian repo? (And of course... we've been waiting for docker-compose itself to be added to those for a really long time :) )

ndeloof:
The install script @ulyssessouza is working on will handle renaming `docker-compose` python into `docker-compose-v1` so that the switch get the expected system setup.

thaJeztah:
For reviewers; the last commit may need some eyes; it has pros/cons (pro: simplifies the dependency between packages, pro/con: the package can be installed without installing the CLI, which can be useful, but theoretically could lead to a user installing only that package (on RPM distros) without having  the docker cli installed.

ndeloof (approved):

thaJeztah:
> The install script @ulyssessouza is working on will handle renaming docker-compose python into docker-compose-v1 so that the switch get the expected system setup.

Sorry, missed that comment: I don't think we can/should do that. The `docker-compose` on a user's machine may have been installed as part of a distro package we don't own (and it's very common for users to have installed those); renaming the binary may cause havoc when users do an `apt-get update` or `apt-get uninstall` (or the RPM equivalent)

i.e., on (for example) ubuntu, running `apt-get install docker-compose` will install the distro-packaged version of docker compose (and docker, as a dependency).

I think the "correct" (and most "natural") approach for the `docker-compose -> docker compose` alias would be a compatibility package (which cannot be named `docker-compose` (see above)), that;

- "conflicts" / "obsoletes" `docker-compose` (the distro package), and thus requires that package to be uninstalled
- "provides" `docker-compose` (? not 100% sure that combination works)
- "requires" `docker-compose-plugin` (the package added in this pull request)
- and which installs the `docker-compose -> docker compose` alias (which could be a script named `docker-compose` with something like `exec docker compose $@`)

fredericdalleau:
wrt to aliasing docker-compose to docker compose, would it make sense/be possible to use the alternatives system?
```
$ docker run --rm -ti debian:buster-slim update-alternatives --get-selections
awk                            auto     /usr/bin/mawk
builtins.7.gz                  auto     /usr/share/man/man7/bash-builtins.7.gz
pager                          auto     /bin/more
rmt                            auto     /usr/sbin/rmt-tar
```

ndeloof (dismissed):

deb/common/control
```diff
@@ -102,6 +102,13 @@ Description: Rootless support for Docker.
   Either VPNKit or slirp4netns (>= 0.4.0) needs to be installed separately.
 Homepage: https://docs.docker.com/engine/security/rootless/
 
+Package: docker-compose-plugin
+Priority: optional
```
justincormack, line 106: This is not optional

thaJeztah: What should it be instead? The dependency is `Enhances: docker-ce-cli`.

tianon: This comes from https://www.debian.org/doc/debian-policy/ch-archive.html#s-priorities, where the most relevant bit is:

> Most Debian packages will have a priority of `optional`. Priority levels other than `optional` are only used for packages that should be included by default in a standard installation of Debian.

tianon: (For a repository that isn't the main distribution repository, `Priority:` doesn't really do much at all.)

deb/common/rules
```diff
@@ -95,6 +110,10 @@ override_dh_install:
 	dh_apparmor --profile-name=docker-ce -pdocker-ce
 
 override_dh_gencontrol:
+	# Use separate version for the compose-plugin package, then generate the other control files as usual
+	# TODO override "Source" field in control as well (to point to compose, as it doesn't match the package name)
+	dh_gencontrol -pdocker-compose-plugin -- -v$${COMPOSE_VERSION#v}~$${DISTRO}-$${SUITE}; \
```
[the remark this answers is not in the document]
tianon: I think you probably just copied it verbatim from the similar line below and corrected the indention but didn't realize it needed a minor change due to the `if` statement :innocent:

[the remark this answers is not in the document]
thaJeztah: Yes, you're right; when editing it, I noticed the lines below, and thought "ah! that's probably why"

rumpl (approved):

PR #551, opened 2021-06-02
Add checks for binary versions set through build-time variables

thaJeztah:
relates to https://github.com/docker/docker-ce-packaging/pull/549

Make sure that these versions are set, and match the expected versions

DEB:

       debian/rules override_dh_auto_test
    make[1]: Entering directory '/root/build-deb'
    ver="$(engine/bundles/dynbinary-daemon/dockerd --version)"; \
        test "$ver" = "Docker version 0.0.0-20210531142756-1c174ced, build 7c6a9484" && echo "PASS: daemon version OK" || echo "FAIL: daemon version ($ver) did not match"
    PASS: daemon version OK
    ver="$(cli/build/docker --version)"; \
        test "$ver" = "Docker version 0.0.0-20210531142756-1c174ced, build 1c174ced" && echo "PASS: cli version OK" || echo "FAIL: cli version ($ver) did not match"
    PASS: cli version OK
    # FIXME: --version currently doesn't work as it makes a connection to the daemon, so using the plugin metadata instead
    ver="$(/usr/libexec/docker/cli-plugins/docker-scan docker-cli-plugin-metadata | awk '{ gsub(/[",:]/,"")}; $1 == "Version" { print $2 }')"; \
        test "$ver" = "v0.8.0" && echo "PASS: docker-scan version OK" || echo "FAIL: docker-scan version ($ver) did not match"
    PASS: docker-scan version OK

RPM:

    Executing(%check): /bin/sh -e /var/tmp/rpm-tmp.SIDNvr
    + umask 022
    + cd /root/rpmbuild/BUILD
    + cd src
    ++ engine/bundles/dynbinary-daemon/dockerd --version
    + ver='Docker version 0.0.0-20210531142756-1c174ced, build 7c6a9484'
    + test 'Docker version 0.0.0-20210531142756-1c174ced, build 7c6a9484' = 'Docker version 0.0.0-20210531142756-1c174ced, build 7c6a9484'
    + echo 'PASS: daemon version OK'
    + exit 0
    PASS: daemon version OK
    ...

    Executing(%check): /bin/sh -e /var/tmp/rpm-tmp.jKzBxw
    + umask 022
    + cd /root/rpmbuild/BUILD
    + cd src
    ++ cli/build/docker --version
    PASS: cli version OK
    + ver='Docker version 0.0.0-20210531142756-1c174ced, build 1c174ced'
[...]

PR #556, opened 2021-07-03
Fix binary version checks masking failure exit code

thaJeztah:
Commit 928a8f2b578a782f902b64d223f06ff423afd736 (https://github.com/docker/docker-ce-packaging/pull/551) added a check for binary set through build-time variables, but I messed up, and forgot to add a non-zero exit code. As a result the exit code was the exit code of the "echo", which would always be successful.

This also revealed a missing check for "target architecture": the scan cli plugin is only built on x86, so the version check should not be performed on other architectures.

relates to https://github.com/docker/docker-ce-packaging/pull/553#issuecomment-873409190

### DIFF
--- a/deb/common/rules
+++ b/deb/common/rules
@@ -30,14 +30,17 @@ override_dh_auto_build:
 
 override_dh_auto_test:
 	ver="$$(engine/bundles/dynbinary-daemon/dockerd --version)"; \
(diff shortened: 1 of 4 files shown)


thaJeztah:
@fredericdalleau ptal 🤗

fredericdalleau (approved):
LGTM

PR #568, opened 2021-08-10
Added support for Rhel 8 and SLES 15 for s390x

nirmannarang:
Added Dockerfile for rhel-8 and sles-15.
Added spec files rpm/SPECS/checkpolicy.spec, rpm/SPECS/policycoreutils.spec for sles-15 support.
Modified rpm/SPECS/docker-ce.spec for adding support for sles-15 with btrfs support.
Modified rpm/Makefile to add support for sles-15 and rhel-8.

Signed-off-by: Nirman Narang <narang@us.ibm.com>


rpm/sles-15/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+ENV RUNC_BUILDTAGS seccomp selinux
+ARG DISTRO
+ARG SUITE
+ENV DISTRO=${DISTRO}
+ENV SUITE=${SUITE}
+ENV CC=gcc
+USER root
+
+
+RUN zypper install -y rpm-build rpmlint pkg-config libsystemd0 systemd-devel selinux-tools wget cmake device-mapper-devel git glibc-devel-static libseccomp-devel libtool libarchive-devel  btrfsprogs libbtrfs-devel lsb-release
```
thaJeztah, line 25: This hard-codes the build dependencies in the dockerfile. I guess this was because SLES doesn't have `yum-builddep`. 

Perhaps instead we can use an alternative, so that we can remove those here, and install them based on the `Buildrequires` in the spec files. @seemethere once wrote a blog about this problem; https://www.terriblecode.com/blog/extracing-rpm-build-dependencies-from-rpm-spec-files/

So something like;

```bash
zipper install $(rpmspec -P ${SPEC_FILE} | grep BuildRequires | cut -d' ' -f2 | xargs)
```

And I think it can even be more simplified, as `rpmspec -q --buildrequires` looks to work (without having to `grep`);


```bash
rpmspec -q --buildrequires SPECS/*.spec 2> /dev/null
git
libtool-ltdl-devel
make
bash
bash
btrfs-progs-devel
ca-certificates
cmake
device-mapper-devel
gcc
git
glibc-static
libarchive
libseccomp-devel
libselinux-devel
libtool
libtool-ltdl-devel
make
pkgconfig
pkgconfig(systemd)
selinux-policy-devel
systemd-devel
tar
which
bash
bash
```

rpm/sles-15/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+
+COPY SPECS /usr/src/packages/SPECS
+COPY rpmbuild/SOURCES/ /usr/src/packages/SOURCES/
+RUN rpmbuild -bb /usr/src/packages/SPECS/policycoreutils.spec
+RUN rpmbuild -bb /usr/src/packages/SPECS/checkpolicy.spec 
+RUN rpm -i /usr/src/packages/RPMS/s390x/policycoreutils-3.2-1.s390x.rpm && rpm -i /usr/src/packages/RPMS/s390x/checkpolicy-3.2.61.1-1.s390x.rpm \
+&& sles_version=$(lsb_release -r | rev | cut -c1) \
+&& opensuse_repo="https://download.opensuse.org/repositories/security:SELinux/SLE_15_SP$sles_version/security:SELinux.repo" \
```
thaJeztah, line 33: Isn't this repository for the same dependencies as the one that are built by the new `.spec` files? (perhaps I misunderstood)

rpm/sles-15/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+ENV CC=gcc
+USER root
+
+
+RUN zypper install -y rpm-build rpmlint pkg-config libsystemd0 systemd-devel selinux-tools wget cmake device-mapper-devel git glibc-devel-static libseccomp-devel libtool libarchive-devel  btrfsprogs libbtrfs-devel lsb-release
+
+COPY SPECS /usr/src/packages/SPECS
+COPY rpmbuild/SOURCES/ /usr/src/packages/SOURCES/
+RUN rpmbuild -bb /usr/src/packages/SPECS/policycoreutils.spec
+RUN rpmbuild -bb /usr/src/packages/SPECS/checkpolicy.spec
```
thaJeztah, line 30: With the `rpmspec --builddeps` approach, I think these steps can be combined with the `docker-c*.spec` step

rpm/SPECS/policycoreutils.spec
```diff
@@ -0,0 +1,31 @@
+Name:           policycoreutils
+Version:        3.2
+Release:        1%{?dist}
+Group:          System Environment/Base
+Summary:        policycoreutils package
+License:        GPLv2
```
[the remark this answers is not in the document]
anujajakhade: We need policycoreutils rpm while building as well as while installing the docker-ce binaries.
Since the version required is not available in repo, we have to build it from source.

rpm/SPECS/policycoreutils.spec
```diff
@@ -0,0 +1,31 @@
+Name:           policycoreutils
```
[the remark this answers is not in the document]
anujajakhade: selinux policy has a hard dependency on policycoreutils. Since the version required is not available in repo, we have to build it from source.

rpm/SPECS/policycoreutils.spec
```diff
@@ -0,0 +1,31 @@
...
+Summary:        policycoreutils package
+License:        GPLv2
+%description
+
+%configure
+
+%install
+wget https://github.com/SELinuxProject/selinux/releases/download/3.2/libsepol-3.2.tar.gz
+zypper install -y gzip make gcc7 flex libbz2-devel libsemanage-devel gettext bison wget tar
+update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-7 7
```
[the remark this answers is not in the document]
anujajakhade: I'll try to include it in spec file.

rpm/sles-15/Dockerfile
anujajakhade: We have enabled this repository for installing selinux packages.
The spec files are used to build the dependencies required by selinux packages like policycoreutils, checkpolicy, etc.
These packages which are available on the repo currently doesn't have s390x support.

rpm/sles-15/Dockerfile
anujajakhade: We need policycoreutils & checkpolicy for installing selinux-tools. So we have to build them before we enable the repo and install selinux packages.

thaJeztah:
I'm a bit lost in the woods currently on the SELinux issue for SLES on s390x. It's a bit hard for me to check all versions, because I don't have a SLES subscription/license, but AFAICS, the current situation is;

- for containerd, there's an exception for SLES, and `container-selinux` is removed as dependency: https://github.com/docker/containerd-packaging/blob/8c11ccb8295aba8a65d55c7fa53b146ba103d67a/rpm/containerd.spec#L55-L59
- for docker engine, we _don't_ have this exception https://github.com/docker/docker-ce-packaging/blob/950efb4c0bdf714fc32cc7017df811cd2633e180/rpm/SPECS/docker-ce.spec#L21
- to provide (the specified version of) the `container-selinux` package, we instruct users to enabled the openSUSE repository (as the SLES repo only has an older version (?)) https://download.opensuse.org/repositories/security:/SELinux/SLE_15_SP3/noarch/

This works, because the `container-selinux` package is a `noarch` package, _but_ it has `policycoreutils` and `selinux-tools` as dependency;

```
curl -O https://ftp.gwdg.de/pub/opensuse/repositories/security:/SELinux/SLE_15_SP3/noarch/container-selinux-2.164.2-16.15.noarch.rpm
rpm -qpR ./container-selinux-2.164.2-16.15.noarch.rpm
warning: ./container-selinux-2.164.2-16.15.noarch.rpm: Header V3 DSA/SHA1 Signature, key ID 93b832ee: NOKEY
/bin/sh
/bin/sh
/bin/sh
/bin/sh
/usr/bin/sed
policycoreutils
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsXz) <= 5.2-1
selinux-policy >= 20210716-112.1
selinux-policy-base >= 20210716
selinux-policy-targeted >= 20210716
selinux-tools
```

And, while the openSUSE repository does provide packages for `x86` and `ppc64le`, it does not support `s390x`; https://download.opensuse.org/repositories/security/openSUSE_Leap_15.3/

You mention:

> We need `policycoreutils` rpm while building as well as while installing the docker-ce binaries. Since the version required is not available in repo, we have to build it from source.

- Do you know _where_ that version requirement is set? (the `container-selinux` package doesn't look to define a specific (minimum) version)
- or is `policycoreutils` not available _at all_ in the default SLES repositories (for s390x)? This page describes packages for s390x, so I _think_ they're available, but perhaps older versions only? https://www.suse.com/support/update/announcement/2019/suse-ru-20191417-1/

I'm "comfortable" with installing a newer version of `container-selinux`, as provided (and maintained) by the openSUSE packagers, but less so with the entire SELinux toolset (besides the GPL Licensing issue). IIUC, this means we'll be upgrading those tools to a version that's not maintained by SUSE (who may be [maintaining patches specific to the distro](https://build.opensuse.org/package/show/openSUSE%3ALeap%3A15.2/policycoreutils)), and will be upgrading to a version that's not supported by them. Upgrading those tools may have more impact than _just_ for their use for containers, and I don't know if running such versions will invalidate support contracts with SUSE (?).

At _least_ it would probably be better to build from the [openSUSE source packages](https://build.opensuse.org/package/show/openSUSE%3ALeap%3A15.2/policycoreutils) (including the patches), instead of directly from upstream, that way the package would match the things we install from the opensuse repository.

Do we know why the openSUSE repository has `ppc64le` but not `s390x` ? Is it because they don't have the hardware for s390x?

anujajakhade:
Hi @thaJeztah,

The version issue is observed when we try to install `selinux-policy`. We need `policycoreutils >= 3.2` & `checkpolicy >= 3.2` for installing selinux-policy.
`policycoreutils` is available in the default [SLES repositories ](https://scc.suse.com/packages?name=SUSE%20Linux%20Enterprise%20Server&version=15.3&arch=s390x&query=policycoreutils&module=)(for s390x), but that is of lower version.

* Attaching the log for reference
```
Hence cannot use binaries available in repo.

#10 340.8 Building repository 'SELinux (SLE_15_SP2)' cache [....done]
#10 340.9 All repositories have been refreshed.
#10 341.2 Refreshing service 'container-suseconnect-zypp'.
#10 341.4 Problem retrieving the repository index file for service 'container-suseconnect-zypp':
#10 341.4 [container-suseconnect-zypp|file:/usr/lib/zypp/plugins/services/container-suseconnect-zypp]
#10 341.4 Warning: Skipping service 'container-suseconnect-zypp' because of the above error.
#10 341.4 Loading repository data...
#10 341.6 Reading installed packages...
#10 341.8 Resolving package dependencies...
#10 341.9 2 Problems:
#10 341.9 Problem: nothing provides 'policycoreutils >= 3.2' needed by the to be installed selinux-policy-20210716-110.1.noarch
#10 341.9 Problem: nothing provides 'checkpolicy >= 3.2' needed by the to be installed selinux-policy-devel-20210716-110.1.noarch
#10 341.9
#10 341.9 Problem: nothing provides 'policycoreutils >= 3.2' needed by the to be installed selinux-policy-20210716-110.1.noarch
#10 341.9  Solution 1: do not install selinux-policy-20210716-110.1.noarch
#10 341.9  Solution 2: break selinux-policy-20210716-110.1.noarch by ignoring some of its dependencies
#10 341.9
#10 341.9 Choose from above solutions by number or skip, retry or cancel [1/2/s/r/c/d/?] (c): c
```
Am not aware why the openSUSE repository doesn't have support for s390x.

rpm/sles-15/Dockerfile
anujajakhade: @thaJeztah 
I have modified the Dockerfile to use `zypper install $(rpmspec -P ${SPEC_FILE} | grep BuildRequires | cut -d' ' -f2 | xargs)`.

rpm/SPECS/policycoreutils.spec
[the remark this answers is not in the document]
anujajakhade: Hi @thaJeztah I have clubbed all the dependencies in `sles_dependencies.spec` file. 
And have added BuildRequires in the respective spec files.

Prabhav-Thali:
Hi @thaJeztah , I have updated the PR with latest changes. Could you please review the same?

Prabhav-Thali:
Hi @thaJeztah , I have updated the PR with latest changes, Which includes updated spec file for policycoreutils-3.4 and also added Dockerfile for RHEL 9. Could you please review?

rpm/rhel-8/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+COPY SPECS /root/rpmbuild/SPECS
+
+# TODO change once we support scan-plugin on other architectures
+RUN \
+  if [ "$(uname -m)" = "x86_64" ]; then \
+    yum-builddep --define '_without_btrfs 1' -y /root/rpmbuild/SPECS/*.spec; \
+  else \
+    yum-builddep --define '_without_btrfs 1' -y /root/rpmbuild/SPECS/docker-c*.spec; \
+    yum-builddep --define '_without_btrfs 1' -y /root/rpmbuild/SPECS/docker-b*.spec; \
+  fi
```
thaJeztah, line 30: Historically, CentOS and RHEL would be "roughly equivalent", while CentOS "stream" no longer has the exact same promises, this still somewhat applies. So I did a compare with the equivalent CentOS Dockerfile; 

- For CentOS-8, we switched from `yum` to `dnf`; I _think_ `dnf` is now considered the preferred (and default) for RHEL 8 and up, and `yum` may even be a symlink to `dnf`, but possibly key import (similar to CentOS) would be needed
- For CentOS, we had to enable additional ("CentOS-Stream-Sources", "CentOS-Stream-PowerTools") repositories; were any additional repos needed for RHEL? (maybe not)


```patch
diff --git a/rpm/rhel-8/Dockerfile b/rpm/centos-8/Dockerfile
index ebe324d..b627d7f 100644
--- a/rpm/rhel-8/Dockerfile
+++ b/rpm/centos-8/Dockerfile
@@ -17,16 +17,27 @@ ARG SUITE
 ENV DISTRO=${DISTRO}
 ENV SUITE=${SUITE}

-RUN yum install -y rpm-build rpmlint libarchive yum-utils
+# In aarch64 (arm64) images, the altarch repo is specified as repository, but
+# failing, so replace the URL.
+RUN if [ -f /etc/yum.repos.d/CentOS-Stream-Sources.repo ]; then sed -i 's/altarch/centos/g' /etc/yum.repos.d/CentOS-Stream-Sources.repo; fi
+
+RUN if [ -f /etc/yum.repos.d/CentOS-Stream-PowerTools.repo ]; then sed -i 's/enabled=0/enabled=1/g' /etc/yum.repos.d/CentOS-Stream-PowerTools.repo; fi
+
+# RHEL8 / CentOS 8 changed behavior and no longer "rpm --import" or
+# "rpmkeys --import"as part of rpm package's %post scriplet. See
+# https://forums.centos.org/viewtopic.php?f=54&t=72574, and
+# https://access.redhat.com/solutions/3720351
+RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial
+RUN dnf install -y rpm-build rpmlint dnf-plugins-core
 COPY SPECS /root/rpmbuild/SPECS

 RUN \
   if [ "$(uname -m)" = "x86_64" ]; then \
-    yum-builddep --define '_without_btrfs 1' -y /root/rpmbuild/SPECS/*.spec; \
+    dnf builddep --define '_without_btrfs 1' -y /root/rpmbuild/SPECS/*.spec; \
   else \
-    yum-builddep --define '_without_btrfs 1' -y /root/rpmbuild/SPECS/docker-c*.spec; \
-    yum-builddep --define '_without_btrfs 1' -y /root/rpmbuild/SPECS/docker-b*.spec; \
+    dnf builddep --define '_without_btrfs 1' -y /root/rpmbuild/SPECS/docker-c*.spec; \
+    dnf builddep --define '_without_btrfs 1' -y /root/rpmbuild/SPECS/docker-b*.spec; \
   fi

 COPY --from=golang /usr/local/go /usr/local/go
```

rpm/rhel-9/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+COPY SPECS /root/rpmbuild/SPECS
+
+# TODO change once we support scan-plugin on other architectures
+RUN \
+  if [ "$(uname -m)" = "x86_64" ]; then \
+    yum-builddep --define '_without_btrfs 1' -y /root/rpmbuild/SPECS/*.spec; \
+  else \
+    yum-builddep --define '_without_btrfs 1' -y /root/rpmbuild/SPECS/docker-c*.spec; \
+    yum-builddep --define '_without_btrfs 1' -y /root/rpmbuild/SPECS/docker-b*.spec; \
+  fi
```
thaJeztah, line 30: Same as the other one; if you're able to look if we need to / should be using `dnf` instead of `yum` ?

rpm/rhel-8/Dockerfile
```diff
@@ -0,0 +1,34 @@
+ARG GO_IMAGE
+ARG DISTRO=rhel
+ARG SUITE=8
+ARG BUILD_IMAGE=dockereng/${DISTRO}:${SUITE}-s390x
```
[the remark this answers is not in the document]
Prabhav-Thali: Hi @thaJeztah, we tried using Red Hat's universal base image for RHEL-8 and RHEL-9. However, Red Hat's official RHEL image does not have a few packages like rpmlint, libtool-ltdl-devel, selinux-policy-devel. To install these packages, repos need to be enabled.
While building docker-ce binaries, we enable the required repos using Red Hat credentials. We can change the image to Red Hat's universal base image. And add a note regarding repos to be enabled.
Please share your thoughts on the same.

rpm/rhel-8/Dockerfile
Prabhav-Thali: > For CentOS-8, we switched from yum to dnf; I think dnf is now considered the preferred (and default) for RHEL 8 and up, and yum may even be a symlink to dnf, but possibly key import (similar to CentOS) would be needed

Yes, sure. We have switched to dnf and added the key import. Will update the PR with same.

> For CentOS, we had to enable additional ("CentOS-Stream-Sources", "CentOS-Stream-PowerTools") repositories; were any additional repos needed for RHEL? (maybe not)

For RHEL 8, `rpmlint` & `libtool-ltdl-devel` requires "rhel-8-for-s390x-appstream-rpms" repo and
`selinux-policy-devel` requires "rhel-8-for-s390x-baseos-rpms"  repo

For RHEL 9, `rpmlint` & `selinux-policy-devel` requires "FTP3install-s390x-9-appstream-os" repo and
`libtool-ltdl-devel` requires "FTP3install-s390x-9-highavailability-os"  repo

To add and use these repositories, credentials are required. As suggested [here](https://github.com/docker/docker-ce-packaging/pull/568#discussion_r1209045190), should we add a note regarding repos to be enabled?

rpm/rhel-9/Dockerfile
Prabhav-Thali: yes, I'll update the PR for this and notify.

rpm/rhel-8/Dockerfile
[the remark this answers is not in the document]
Prabhav-Thali: Hi @thaJeztah, Any thoughts on the above?

PR #571, opened 2021-08-27
Require `dbus-user-session` for rootless

AkihiroSuda:
On Debian, `dbus-user-session` is not installed by default.

The lack of `dbus-user-session` results in a cryptic error on rootless+cgroup2+systemd: `read unix @->/run/systemd/private: read: connection reset by peer: unknown.`

ref: moby/moby#42793


thaJeztah (approved):
LGTM

PR #598, opened 2021-11-16
[master] update buildx to v0.7.0

thaJeztah:
release notes: https://github.com/docker/buildx/releases/tag/v0.7.0

### DIFF
--- a/plugins/buildx.installer
+++ b/plugins/buildx.installer
@@ -6,7 +6,7 @@ source "$(dirname "$0")/.common"
 PKG=github.com/docker/buildx
 GOPATH=$(go env GOPATH)
 REPO=https://${PKG}.git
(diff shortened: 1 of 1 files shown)


crazy-max:
Buildx requires Go 1.17 so we might need to first upgrade Go 1.17 on docker/cli https://github.com/docker/cli/pull/3258 and after that update Go here:

https://github.com/docker/docker-ce-packaging/blob/8bbb124d84a94765592fcb4aeef66982daf2d69d/common.mk#L19

thaJeztah:
It's no longer buildable with go 1.16?

crazy-max:
> It's no longer buildable with go 1.16?

Ah yes it can we have not changed the `go.mod` to keep backward compt so it should be ok.

crazy-max (approved):

PR #633, opened 2022-03-04
install recent Git version on CentOS 7

crazy-max:
follow-up https://github.com/docker/docker-ce-packaging/pull/631#issuecomment-1059363763

see also https://github.com/docker/docker-ce-packaging/pull/620#discussion_r793943247

Signed-off-by: CrazyMax <crazy-max@users.noreply.github.com>

PR #446, opened 2020-03-25
CentOS: fix failing repository on aarch64 (arm64)

thaJeztah:
This repository is only specified in the aarch64 (arm64) images,
and therefore was not caught in CI.

Bringing back the fix that was previously there, to prevent failures:

    #9 [stage-1 4/6] RUN yum-builddep -y /root/rpmbuild/SPECS/*.spec
    #9 0.982 Loaded plugins: fastestmirror, ovl
    #9 1.177 Enabling base-source repository
    #9 1.177 Enabling extras-source repository
    #9 1.178 Enabling updates-source repository
    #9 1.179 Loading mirror speeds from cached hostfile
    #9 1.181  * base: d36uatko69830t.cloudfront.net
    #9 1.182  * extras: d36uatko69830t.cloudfront.net
    #9 1.182  * updates: d36uatko69830t.cloudfront.net
    #9 2.272 http://vault.centos.org/altarch/7/extras/Source/repodata/repomd.xml: [Errno 14] HTTP Error 404 - Not Found
    #9 2.272 Trying other mirror.
    #9 2.273 To address this issue please refer to the below wiki article

PR #416, opened 2020-01-16
Add Support for Centos-8

arkodg:

PR #634, opened 2022-03-07
[master] Build for CentOS 9.

Romain-Geissler-1A:
This shall be backported to branch 20.10 too.

### DIFF
--- a/rpm/SPECS/docker-ce.spec
+++ b/rpm/SPECS/docker-ce.spec
@@ -1,7 +1,9 @@
 %global debug_package %{nil}
 
 # BTRFS is enabled by default, but can be disabled by defining _without_btrfs
-%{!?_with_btrfs: %{!?_without_btrfs: %define _with_btrfs 1}}
+%if %{undefined _with_btrfs} && %{undefined _without_btrfs}
+%define _with_btrfs 1
+%endif
 
 Name: docker-ce
 Version: %{_version}
@@ -22,7 +24,10 @@ Requires: container-selinux >= 2:2.74
 Requires: libseccomp >= 2.3
 Requires: systemd
 Requires: iptables
+%if %{undefined rhel} || 0%{?rhel} < 9
+# Libcgroup is no longer available in RHEL/CentOS >= 9 distros.
 Requires: libcgroup
+%endif
 Requires: containerd.io >= 1.4.1
 Requires: tar
 Requires: xz
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,6 +5,7 @@ def branch = env.CHANGE_TARGET ?: env.BRANCH_NAME
 def pkgs = [
     [target: "centos-7",                 image: "centos:7",                               arches: ["amd64", "aarch64"]],          // (EOL: June 30, 2024)
     [target: "centos-8",                 image: "quay.io/centos/centos:stream8",          arches: ["amd64", "aarch64"]],
+    [target: "centos-9",                 image: "quay.io/centos/centos:stream9",          arches: ["amd64", "aarch64"]],
     [target: "debian-buster",            image: "debian:buster",                          arches: ["amd64", "aarch64", "armhf"]], // Debian 10 (EOL: 2024)
     [target: "debian-bullseye",          image: "debian:bullseye",                        arches: ["amd64", "aarch64", "armhf"]], // Debian 11 (Next stable)
     [target: "fedora-34",                image: "fedora:34",                              arches: ["amd64", "aarch64"]],          // EOL: May 17, 2022
--- a/rpm/Makefile
+++ b/rpm/Makefile
@@ -57,7 +57,7 @@ RUN?=docker run --rm \
 	rpmbuild-$@/$(ARCH) $(RPMBUILD_FLAGS)
 
 FEDORA_RELEASES ?= fedora-36 fedora-35 fedora-34
-CENTOS_RELEASES ?= centos-7 centos-8
+CENTOS_RELEASES ?= centos-7 centos-8 centos-9
 ifeq ($(ARCH),s390x)
 RHEL_RELEASES ?= rhel-7
 else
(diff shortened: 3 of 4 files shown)


Romain-Geissler-1A:
Note: this will required #633 to be merged first.

Romain-Geissler-1A:
Note: when this will be merged and https://github.com/docker/containerd-packaging/pull/270 will be merged/delivered, I have the local changes to build for CentOS 9 too.

rpm/SPECS/docker-ce.spec
```diff
@@ -22,7 +22,10 @@ Requires: container-selinux >= 2:2.74
 Requires: libseccomp >= 2.3
 Requires: systemd
 Requires: iptables
+%if ! 0%{?rhel}
```
thaJeztah, line 25: Should we make this conditional based on `rhel` (/centos) version? If it's not deprecated in rhel7 and centos7, (or 8) we could make it something like;

```suggestion
%if 0%{?rhel} == 7 || 0%{?rhel} == 8
```

thaJeztah: Using straight (`==`) comparisons to account for other distros who would like not have the `rhel` macro (but double check if my suggestion is correct 😅)

Romain-Geissler-1A: Well, libcgroup is deprecated in RHEL 7 already: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/resource_management_guide/chap-using_libcgroup_tools

> The libcgroup package, which was the main tool for cgroup management in previous versions of Red Hat Enterprise Linux, is now deprecated. To avoid conflicts, do not use libcgroup tools for default resource controllers (listed in [Available Controllers in Red Hat Enterprise Linux 7](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/resource_management_guide/br-Resource_Controllers_in_Linux_Kernel#itemlist-Available_Controllers_in_Red_Hat_Enterprise_Linux_7)) that are now an exclusive domain of systemd. This leaves a limited space for applying libcgroup tools, use it only when you need to manage controllers not currently supported by systemd, such as net_prio.

thaJeztah: The devil can be in the detail

> This leaves a limited space for applying libcgroup tools, use it only when you need to manage controllers not currently supported by systemd, such as net_prio.

so I think we should consider keeping it (for now) for the versions that still carry the package

Romain-Geissler-1A: Ok, then. I used the expression `%if 0%{?rhel} <= 8` does it fly for you ?

Romain-Geissler-1A:
Note that another prerequisite to building for RHEL/CentOS 9 is considering something like https://github.com/docker/containerd-packaging/pull/270 (which is failing the CI, but since external contributors can't see the Jenkins logs, I don't know it's failing).

Romain-Geissler-1A:
Ping

Romain-Geissler-1A:
Ping

Romain-Geissler-1A:
Ping

Romain-Geissler-1A:
@thaJeztah Pull request rebased and updated to adopt your style as in https://github.com/docker/containerd-packaging/pull/283. I have also modified the pull request purpose, now it removes libcgroup dependency for RHEL >= 9 and also builds a package for CentOS 9 at the same time. Most likely the CI will remain red until a containerd package for CentOS 9 is delivered in the official docker repositories.

rpm/centos-8/Dockerfile
```diff
@@ -18,27 +18,23 @@ ARG SUITE
 ENV DISTRO=${DISTRO}
 ENV SUITE=${SUITE}
 
-# In aarch64 (arm64) images, the altarch repo is specified as repository, but
-# failing, so replace the URL.
-RUN if [ -f /etc/yum.repos.d/CentOS-Stream-Sources.repo ]; then sed -i 's/altarch/centos/g' /etc/yum.repos.d/CentOS-Stream-Sources.repo; fi
-
```
thaJeztah, line 24: This one was an odd one; I know we added it at some point, then we removed it (because it appeared it didn't have the `altarch` anymore); https://github.com/docker/docker-ce-packaging/pull/416#discussion_r367891915

But later, we had to add it back, because for some reason, our pipeline started failing again on this; see https://github.com/docker/docker-ce-packaging/pull/446

Could it be these can be set dynamically, depending on what mirror is hit?

thaJeztah, line 24: This one was an odd one; I know we added it at some point, then we removed it (because it appeared it didn't have the `altarch` anymore); https://github.com/docker/docker-ce-packaging/pull/416#discussion_r367891915

But later, we had to add it back, because for some reason, our pipeline started failing again on this; see https://github.com/docker/docker-ce-packaging/pull/446

Could it be these can be set dynamically, depending on what mirror is hit?

Romain-Geissler-1A: I thought this was some old hack no longer necessary. But if you say it's still needed, I will put it back.

thaJeztah: The honest answer: "I don't know!" 😂 (I also thought it was redundant, and then... all of a sudden, it failed in one of our internal pipelines)

Perhaps we should keep it "just to be safe" (it looks like it doesn't hurt to have it, and would be a no-op if it's not needed).

thaJeztah:
Overall; changes look good to me; I'm running a build locally, as CI doesn't pick up changes in Jenkinsfiles for users without write-access

thaJeztah:
Working on a (temporary) fix for failing CI; looks like someone's vanity URL went down; https://github.com/docker/docker-ce-packaging/pull/684

thaJeztah:
I kicked CI, but looks like GitHub actions doesn't do a merge before running again (whereas jenkins does).

I tested this locally, and it looks to be working as expected

thaJeztah (approved):
LGTM, thanks!

PR #415, opened 2020-01-14
Include plugins in static build

HeroCC:
fixes https://github.com/docker/for-linux/issues/1066

Docker documentation states that plugins like buildx and app are included in distributions as of v19.13 (https://github.com/docker/docker-ce-packaging/pull/309).   However, this doesn't include the static builds. Downstream, people expect these to be included and may be confused to learn that the static build doesn't include them: https://github.com/docker-library/docker/issues/156. Including the plugins would make the builds more consistent.

This is my first contribution to the docker project, so apologies if this breaks anything 😅

PR #654, opened 2022-03-25
[master] add docker-buildx-plugin package (deb, rpm, static)

### LINKED ISSUE #618: Combining static tarball and default cli-plugins
With https://github.com/docker/cli/pull/3314 updates, there is an open question if we should do something with the static binary bundle. With that PR, in order to build with buildkit, buildx binary is needed. The packages already contain buildx, but the static tarball at https://download.docker.com/linux/static/stable/x86_64/ does not.

We should also take into account that more plugins like this will appear in the future. Compose rewrite in go is also a plugin, and Docker Desktop already ships with `docker-scan` with more coming in the future. The immediate concern for the 21.x release is `buildx` though.

The options are (comment if I missed any):
- Add the buildx to the current static tarball. It would be the 10th binary there. Probably the easiest solution.
- Create a special `cli-plugins` tarball. Currently containing buildx. Future additions would be added there.
- Create a `cli` tarball that contains `cli` and additional client-side binaries. This has been a requested outcome independently from this work to match the way packages are separated. If we want to do this eventually, then we might as well do these changes together now.
- Add static links from https://download.docker.com/linux/static to https://github.com/docker/buildx/releases https://github.com/docker/compose/releases and https://github.com/docker/scan-cli-plugin/releases . These already point to static binaries for all architectures and can be used directly when someone wants to `curl` a binary. We would not update these links with new docker releases.

There is also an option to not do anything and consider static binaries an advanced use case. When users can figure out how to download and configure them, they can figure out the same for other binaries as well.

Personally, I don't have a preference and would be ok with any of the options. We just need to make a conscious decision.

crazy-max:
fixes #618
closes #415

for the upcoming major docker release we want to make sure buildx is part of its own package like scan and compose. currently, buildx is part of the docker-ce-cli package.

this also creates the static tarball.

last commit removes the plugin installers logic as it was only used for buildx and should not be used in the future.

in a follow-up we want to fetch plugins packages from upstream repositories instead of rebuilding them everytime. see https://github.com/docker/buildx/pull/840

```console
$ VERSION=22.04.0 make debian-bullseye
$ VERSION=22.04.0 make centos-7
$ VERSION=22.04.0 make DOCKER_BUILD_PKGS='static-linux' static
```


rpm/SPECS/docker-ce-cli.spec
```diff
@@ -109,7 +88,6 @@ done
 %files
 %doc build-docs/LICENSE build-docs/MAINTAINERS build-docs/NOTICE build-docs/README.md
 %{_bindir}/docker
-%{_libexecdir}/docker/cli-plugins/*
```
crazy-max, line 112: needs to remove that line to fix: https://ci-next.docker.com/public/blue/organizations/jenkins/docker-ce-packaging/detail/PR-654/2/pipeline#step-893-log-356

```
PASS: cli version OK

Processing files: docker-ce-cli-0.0.0.20220325091330.87b8e57-0.el7.aarch64

error: File not found by glob: /root/rpmbuild/BUILDROOT/docker-ce-cli-0.0.0.20220325091330.87b8e57-0.el7.aarch64/usr/libexec/docker/cli-plugins/*

Executing(%doc): /bin/sh -e /var/tmp/rpm-tmp.hH1z6a
```

we don't ship plugins inside docker-ce-cli pkg anymore.

crazy-max:
I'm not able to build locally I'm not sure why:

```
$ make REF=master debian-bullseye
...
# docker-ce-rootless-extras install
install -D -m 0755 /usr/local/bin/rootlesskit debian/docker-ce-rootless-extras/usr/bin/rootlesskit
install -D -m 0755 /usr/local/bin/rootlesskit-docker-proxy debian/docker-ce-rootless-extras/usr/bin/rootlesskit-docker-proxy
install -D -m 0755 engine/contrib/dockerd-rootless.sh debian/docker-ce-rootless-extras/usr/bin/dockerd-rootless.sh
install -D -m 0755 engine/contrib/dockerd-rootless-setuptool.sh debian/docker-ce-rootless-extras/usr/bin/dockerd-rootless-setuptool.sh
# TODO: how can we install vpnkit?
make[1]: Leaving directory '/root/build-deb'
   debian/rules override_dh_install
make[1]: Entering directory '/root/build-deb'
dh_install
# TODO Can we do this from within our container?
make[1]: Leaving directory '/root/build-deb'
   dh_installdocs
cli/README.md: 3: Syntax error: word unexpected (expecting ")")
dh_installdocs: warning: debian/docs is marked executable but does not appear to an executable config.
dh_installdocs: warning:
dh_installdocs: warning: If debian/docs is intended to be an executable config file, please ensure it can
dh_installdocs: warning: be run as a stand-alone script/program (e.g. "./debian/docs")
dh_installdocs: warning: Otherwise, please remove the executable bit from the file (e.g. chmod -x "debian/docs")
dh_installdocs: warning:
dh_installdocs: warning: Please see "Executable debhelper config files" in debhelper(7) for more information.
dh_installdocs: warning:
dh_installdocs: error: debian/docs (executable config) returned exit code 2
make: *** [debian/rules:127: binary] Error 25
dpkg-buildpackage: error: debian/rules binary subprocess returned exit status 2
make[1]: *** [Makefile:78: debian-bullseye] Error 2
```

Seems fine on CI: https://ci-next.docker.com/public/blue/organizations/jenkins/docker-ce-packaging/detail/PR-654/3/pipeline

Any idea @thaJeztah?

crazy-max:
As discussed about https://github.com/docker/docker-ce-packaging/pull/654#issuecomment-1078964391, it's an issue with my local env.

crazy-max:
Comparing pkg generated on https://github.com/docker/buildx/pull/840 and with this PR:

```console
# with https://github.com/docker/buildx/pull/840
$ dpkg-deb -I docker-buildx-plugin_0.7.0~89-g4a49a60f.m-debian11_amd64.deb
 new Debian package, version 2.0.
 size 12365532 bytes: control archive=424 bytes.
       1 bytes,     0 lines      conffiles
     308 bytes,     9 lines      control
      81 bytes,     1 lines      md5sums
 Package: docker-buildx-plugin
 Version: 0:0.7.0-debian11~89-g4a49a60f.m
 Section: default
 Priority: optional
 Architecture: amd64
 Maintainer: Docker <support@docker.com>
 Installed-Size: 46516
 Homepage: https://github.com/docker/buildx
 Description: Docker Buildx plugin extends build capabilities with BuildKit.
```

```console
# with docker-ce-packaing
$ dpkg-deb -I docker-buildx-plugin_0.8.1~debian-bullseye_amd64.deb
 new Debian package, version 2.0.
 size 23107212 bytes: control archive=568 bytes.
     352 bytes,    11 lines      control
     168 bytes,     2 lines      md5sums
 Package: docker-buildx-plugin
 Source: docker-ce (5:0.0.0-20220325091330-87b8e57-0~debian-bullseye)
 Version: 0.8.1~debian-bullseye
 Architecture: amd64
 Maintainer: Docker <support@docker.com>
 Installed-Size: 67146
 Enhances: docker-ce-cli
 Section: admin
 Priority: optional
 Homepage: https://github.com/docker/buildx
 Description: Docker Buildx cli plugin.
```

crazy-max:
Looks like docker-ce-cli pkg also installs `docker-app`, not sure where it comes from:

```console
$ dpkg -c docker-ce-cli_20.10.14~3-0~debian-bullseye_amd64.deb
drwxr-xr-x root/root         0 2022-03-24 01:46 ./
drwxr-xr-x root/root         0 2022-03-24 01:46 ./usr/
drwxr-xr-x root/root         0 2022-03-24 01:46 ./usr/bin/
-rwxr-xr-x root/root  61885504 2022-03-24 01:46 ./usr/bin/docker
drwxr-xr-x root/root         0 2022-03-24 01:46 ./usr/libexec/
drwxr-xr-x root/root         0 2022-03-24 01:46 ./usr/libexec/docker/
drwxr-xr-x root/root         0 2022-03-24 01:46 ./usr/libexec/docker/cli-plugins/
-rwxr-xr-x root/root  33637568 2022-03-24 01:46 ./usr/libexec/docker/cli-plugins/docker-app
-rwxr-xr-x root/root  68787979 2022-03-24 01:46 ./usr/libexec/docker/cli-plugins/docker-buildx
...
```

thaJeztah:
yes 20.10.14 still had the `docker-app` plugin included; it's removed in the master branch

crazy-max:
@thaJeztah This also adds the static package for buildx:

```console
$ VERSION=22.04.0 make DOCKER_BUILD_PKGS='static-linux' static
...
$ ls static/build/linux/
docker  docker-22.04.0.tgz  docker-buildx-plugin-0.8.1.tgz  docker-rootless-extras  docker-rootless-extras-22.04.0.tgz
```

I made some tests to check everything is ok with rpm and deb packages because we are now splitting packages (`docker-ce-cli` with `docker-buildx-plugin`) so there is an implicit conflict with `/usr/libexec/docker/cli-plugins/docker-buildx`:

```console
$ VERSION=22.04.0 make centos-7
$ cd rpm/rpmbuild/
$ docker run --rm -it -v $(pwd)/centos-7/RPMS/x86_64:/pkgs centos:7 bash
```

```console
[root@cb4fce82d0ec /]# yum install -y yum-utils
[root@cb4fce82d0ec /]# yum-config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo
[root@cb4fce82d0ec /]# yum install -y docker-ce-cli
...
Dependencies Resolved

========================================================================================================================================================================================================== Package                                             Arch                                    Version                                              Repository                                         Size 
==========================================================================================================================================================================================================Installing:
 docker-ce-cli                                       x86_64                                  1:20.10.14-3.el7                                     docker-ce-stable                                   30 M 
Installing for dependencies:
 docker-scan-plugin                                  x86_64                                  0.17.0-3.el7                                         docker-ce-stable                                  3.7 M 

Transaction Summary
==========================================================================================================================================================================================================Install  1 Package (+1 Dependent package)

Total download size: 34 M
Installed size: 158 M
Downloading packages:
warning: /var/cache/yum/x86_64/7/docker-ce-stable/packages/docker-scan-plugin-0.17.0-3.el7.x86_64.rpm: Header V4 RSA/SHA512 Signature, key ID 621e9f35: NOKEY
Public key for docker-scan-plugin-0.17.0-3.el7.x86_64.rpm is not installed
(1/2): docker-scan-plugin-0.17.0-3.el7.x86_64.rpm                                                                                                                                  | 3.7 MB  00:00:00     
(2/2): docker-ce-cli-20.10.14-3.el7.x86_64.rpm                                                                                                                                     |  30 MB  00:00:00     
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------Total                                                                                                                                                                      50 MB/s |  34 MB  00:00:00     
Retrieving key from https://download.docker.com/linux/centos/gpg
Importing GPG key 0x621E9F35:
 Userid     : "Docker Release (CE rpm) <docker@docker.com>"
 Fingerprint: 060a 61c5 1b55 8a7f 742b 77aa c52f eb6b 621e 9f35
 From       : https://download.docker.com/linux/centos/gpg
Running transaction check
Running transaction test
Transaction test succeeded
Running transaction
  Installing : 1:docker-ce-cli-20.10.14-3.el7.x86_64                                                                                                                                                  1/2 
  Installing : docker-scan-plugin-0.17.0-3.el7.x86_64                                                                                                                                                 2/2 
  Verifying  : docker-scan-plugin-0.17.0-3.el7.x86_64                                                                                                                                                 1/2 
  Verifying  : 1:docker-ce-cli-20.10.14-3.el7.x86_64                                                                                                                                                  2/2 

Installed:
  docker-ce-cli.x86_64 1:20.10.14-3.el7

Dependency Installed:
  docker-scan-plugin.x86_64 0:0.17.0-3.el7

Complete!
```

```console
[root@cb4fce82d0ec /]# yum install -y createrepo
[root@cb4fce82d0ec /]# createrepo /pkgs
[root@cb4fce82d0ec /]# cat > /etc/yum.repos.d/docker-test.repo <<EOL
[docker-test]
name=DockerTest
baseurl=file:///pkgs
enabled=1
gpgcheck=0
EOL
[root@cb4fce82d0ec /]# yum update -y docker-ce-cli
...
Dependencies Resolved

========================================================================================================================================================================================================== Package                                                Arch                                     Version                                              Repository                                     Size 
==========================================================================================================================================================================================================Updating:
 docker-ce-cli                                          x86_64                                   1:22.04.0-3.el7                                      docker-test                                   7.4 M 
Installing for dependencies:
 docker-buildx-plugin                                   x86_64                                   0.8.1-3.el7                                          docker-test                                    11 M 

Transaction Summary
==========================================================================================================================================================================================================Install             ( 1 Dependent package)
Upgrade  1 Package

Total download size: 19 M
Downloading packages:
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------Total                                                                                                                                                                     1.2 GB/s |  19 MB  00:00:00     
Running transaction check
Running transaction test
Transaction test succeeded
Running transaction
  Installing : docker-buildx-plugin-0.8.1-3.el7.x86_64                                                                                                                                                1/3 
  Updating   : 1:docker-ce-cli-22.04.0-3.el7.x86_64                                                                                                                                                   2/3 
  Cleanup    : 1:docker-ce-cli-20.10.14-3.el7.x86_64                                                                                                                                                  3/3 
  Verifying  : 1:docker-ce-cli-22.04.0-3.el7.x86_64                                                                                                                                                   1/3 
  Verifying  : docker-buildx-plugin-0.8.1-3.el7.x86_64                                                                                                                                                2/3 
  Verifying  : 1:docker-ce-cli-20.10.14-3.el7.x86_64                                                                                                                                                  3/3 

Dependency Installed:
  docker-buildx-plugin.x86_64 0:0.8.1-3.el7

Updated:
  docker-ce-cli.x86_64 1:22.04.0-3.el7

Complete!
```

thaJeztah:
hmm interesting; looks like something is using the wrong go version perhaps?? (or do we still have a go version hard-coded? perhaps we have in this repo let me check)

```
./scripts/vendor update
+ go mod tidy -compat=1.17 -modfile=vendor.mod
flag provided but not defined: -compat
usage: go mod tidy [-e] [-v]

**Comment by thaJeztah** [comment]
opened https://github.com/docker/docker-ce-packaging/pull/655

File: static/Makefile [hunk 2]
```diff
@@ -40,12 +42,14 @@ clean: ## remove build artifacts
...
 .PHONY: static-linux
-static-linux: static-cli static-engine ## create tgz with linux x86_64 client and server
+static-linux: static-cli static-engine static-buildx-plugin ## create tgz
 	mkdir -p build/linux/docker
 	cp $(CLI_DIR)/build/docker build/linux/docker/
 	for f in dockerd containerd ctr containerd-shim containerd-shim-runc-v2 docker-init docker-proxy runc; do \
 		cp -L $(ENGINE_DIR)/bundles/binary-daemon/$$f build/linux/docker/$$f; \
 	done
+	mkdir -p build/linux/docker/cli-plugins
+	cp $(BUILDX_DIR)/bin/docker-buildx build/linux/docker/cli-plugins/
```
crazy-max, line 52: As discussed, not sure if we should include buildx in the docker-ce-cli static pkg even if it's a required component now.

crazy-max: Removed for now

deb/common/control
```diff
@@ -102,6 +103,14 @@ Description: Rootless support for Docker.
   Either VPNKit or slirp4netns (>= 0.4.0) needs to be installed separately.
 Homepage: https://docs.docker.com/engine/security/rootless/
 
+Package: docker-buildx-plugin
+Priority: optional
+Replaces: docker-ce-cli
```
thaJeztah, line 108: I initially was confused by the `Replaces:` here, as I thought it would mark this package to replace the whole `docker-ce-cli` package, but @crazy-max explained this also can be used to replace _some files_ (given, it's a bit "fuzzy", as it doesn't require speciyfing _which_ files), from the [The Debian Administrator's Handbook](https://debian-handbook.info/browse/stable/sect.package-meta-information.html#id-1.8.6.7.14):


> ### 5.2.1.5. Replacing Files: The Replaces Field

> The Replaces field indicates that the package contains files that are also present in another package, but that the package is legitimately entitled to replace them. Without this specification, dpkg fails, stating that it can not overwrite the files of another package (technically, it is possible to force it to do so with the --force-overwrite option, but that is not considered standard operation). This allows identification of potential problems and requires the maintainer to study the matter prior to choosing whether to add such a field.
>
>The use of this field is justified when package names change or when a package is included in another. This also happens when the maintainer decides to distribute files differently among various binary packages produced from the same source package: a replaced file no longer belongs to the old package, but only to the new one.
>
> If all of the files in an installed package have been replaced, the package is considered to be removed. Finally, this field also encourages dpkg to remove the replaced package where there is a conflict.

The last paragraph outlines our situation

What we probably _could_ consider doing is (if supported) to include the version here, e.g.

```suggestion
Replaces: docker-ce-cli (<< 22.04.0)
```

But we need to check if it likes CalVer; instead, probably this would work;

```suggestion
Replaces: docker-ce-cli (<= 21.10~)
```

deb/common/control
```diff
@@ -61,7 +61,8 @@ Description: Docker: the open-source application container engine
 
 Package: docker-ce-cli
 Architecture: linux-any
-Depends: ${shlibs:Depends}
+Depends: docker-buildx-plugin,
+         ${shlibs:Depends}
 # TODO change once we support scan-plugin on other architectures
 Recommends: docker-scan-plugin [amd64]
```
thaJeztah, line 67: I'm still thinking if this should be `Recommends` to allow for "small" installations that are not used to build images (e.g. production situations). I realise the `buildx` plugin is required to perform `docker build`, but `docker run` (and anything else) can function without it. We already "control" that situation in the main CLI (produce an error if the component is missing when someone attempts to do a `docker build`); `Recommend` in itself is already a quite "strong" dependency, and by default will be installed. From [the handbook](https://debian-handbook.info/browse/stable/sect.package-meta-information.html#id-1.8.6.7.10.10);

> (...) the “recommended” dependencies, the most important, considerably improve the functionality offered by the package but are not indispensable to its operation.

And

> (...) You should always install the “recommended” packages, unless you know exactly why you do not need them. This is now also the default for APT unless configured otherwise.

deb/common/rules
```diff
@@ -110,6 +107,10 @@ override_dh_install:
 	dh_apparmor --profile-name=docker-ce -pdocker-ce
 
 override_dh_gencontrol:
+	# Use separate version for the buildx-plugin package, then generate the other control files as usual
+	# TODO override "Source" field in control as well (to point to buildx, as it doesn't match the package name)
```
thaJeztah, line 111: We should probably (in a follow-up) duplicate the current `deb/common` directory for each of the packages we build, and split the control file accordingly. That will be a bit of work though (also splitting make targets etc etc)

Ideally, have separate stages in the Dockerfile to collect the source-files (and dependencies) for each, also to get out of the current situation where we bind-mount things and build packages in a `docker run` (which is horrible, because not only do we lose (e.g.) modules cache, but also currently requires some hacks `chown` / `chmod` to be done to adjust file permissions for cleaning up 😞

rpm/centos-8/Dockerfile
```diff
@@ -38,6 +38,7 @@ RUN \
     yum-builddep --define '_without_btrfs 1' -y /root/rpmbuild/SPECS/*.spec; \
   else \
     yum-builddep --define '_without_btrfs 1' -y /root/rpmbuild/SPECS/docker-c*.spec; \
+    yum-builddep --define '_without_btrfs 1' -y /root/rpmbuild/SPECS/docker-b*.spec; \
```
thaJeztah, line 41: We probably don't need this line; I think this was specifically using `docker-c` so that we only defined this for the `docker-ce-` packages (the engine uses the `btrfs` build-tag); probably the existing one should even be changed to `docker-ce.spec`, as `docker-ce-cli.spec` likely doesn't need it either, but will need to check what we need there.

deb/common/control
crazy-max: Unfortunately I tried and it doesn't work with CalVer

deb/common/control
crazy-max: Yes agree, will put buildx that to Recommends instead.

deb/common/rules
crazy-max: Yeah I already have a draft branch for that :eyes:

deb/common/control
crazy-max: About relationships see also https://www.debian.org/doc/debian-policy/ch-relationships.html#overwriting-files-and-replacing-packages-replaces

rpm/centos-8/Dockerfile
crazy-max: Yes let's do that in a follow-up.

thaJeztah (dismissed):
LGTM

thaJeztah:
oh! you need to update the fedora-36 dockerfile now as well

thaJeztah (approved):
LGTM

PR #656, opened 2022-03-26
[master] makefile: add PACKAGER_NAME common variable

crazy-max:
follow-up https://github.com/docker/cli/pull/3310 and https://github.com/moby/moby/pull/43431

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -15,6 +15,7 @@ ARCHES:=amd64 aarch64 armhf armel s390x ppc64le
 BUILDTIME=$(shell date -u -d "@$${SOURCE_DATE_EPOCH:-$$(date +%s)}" --rfc-3339 ns 2> /dev/null | sed -e 's/ /T/')
 CHOWN:=docker run --rm -v $(CURDIR):/v -w /v alpine chown
 DEFAULT_PRODUCT_LICENSE:=Community Engine
+PACKAGER_NAME?=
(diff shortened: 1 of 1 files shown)


common.mk
```diff
@@ -15,6 +15,7 @@ ARCHES:=amd64 aarch64 armhf armel s390x ppc64le
 BUILDTIME=$(shell date -u -d "@$${SOURCE_DATE_EPOCH:-$$(date +%s)}" --rfc-3339 ns 2> /dev/null | sed -e 's/ /T/')
 CHOWN:=docker run --rm -v $(CURDIR):/v -w /v alpine chown
 DEFAULT_PRODUCT_LICENSE:=Community Engine
+COMPANY_NAME:=Docker, Inc.
```
thaJeztah, line 18: perhaps it should be empty by default (but we can set it in our internal pipeline)
that way, someone using the code from this repository doesn't set docker's name as packager

thaJeztah:
I just realised we discussed naming of this variable on Slack, but already used `COMPANY_NAME` on the cli. Do you think `PACKAGER_NAME` would be more clear on intent (and possibly better reusable also for rpm/deb)? I opened a quick PR on the cli to change the variable there; https://github.com/docker/cli/pull/3500

thaJeztah (approved):
LGTM

PR #657, opened 2022-03-27
[master] rpm: fix circular dependency between scan-cli-plugin and docker-ce-cli

thaJeztah:
This fixes a problem when installing the packages from local files instead of
from the package repository. The current packages had a strong dependency set
in both directions (`docker-ce-cli` requires `docker-scan-plugin`, and vice-
versa). This caused a circular dependency, which could not be resolved when
trying to install the packages offline (from downloaded rpm files);

    yum install ./docker-scan-plugin-0.17.0-3.fc35.x86_64.rpm
    Last metadata expiration check: 1:09:40 ago on Tue Mar  8 08:30:47 2022.
    Error:
    Problem: conflicting requests
    - nothing provides docker-ce-cli needed by docker-scan-plugin-0.17.0-3.fc35.x86_64
      (try to add '--skip-broken' to skip uninstallable packages)

    yum install ./docker-ce-cli-20.10.13-3.fc35.x86_64.rpm
    Last metadata expiration check: 1:10:23 ago on Tue Mar  8 08:30:47 2022.
    Error:
    Problem: conflicting requests
    - nothing provides docker-scan-plugin(x86-64) needed by docker-ce-cli-1:20.10.13-3.fc35.x86_64
      (try to add '--skip-broken' to skip uninstallable packages)

This patch

- changes the `docker-scan-plugin` to `Enhances
- changes the `docker-ce-cli` package to mark the scan plugin as `Recommends`,
  with the exception of CentOS/RHEL 7, which do not yet support weak dependencies.
  For those, we continue to use `Requires`.

The `Recommends` dependency should be installed by default, but users can opt-out
by passing `--setopt=install_weak_deps=False` to `dnf`.

### DIFF
--- a/rpm/SPECS/docker-ce-cli.spec
+++ b/rpm/SPECS/docker-ce-cli.spec
@@ -18,7 +18,15 @@ Requires: /bin/sh
 Requires: /usr/sbin/groupadd
 # TODO change once we support scan-plugin on other architectures
 %ifarch x86_64
+# CentOS 7 and RHEL 7 do not yet support weak dependencies
+#
+# Note that we're not using <= 7 here, to account for other RPM distros, such
+# as Fedora, which would not have the rhel macro set (so default to 0).
+%if 0%{?rhel} == 7
 Requires: docker-scan-plugin(x86-64)
+%else
+Recommends: docker-scan-plugin(x86-64)
+%endif
 %endif
 
 BuildRequires: make
--- a/rpm/SPECS/docker-scan-plugin.spec
+++ b/rpm/SPECS/docker-scan-plugin.spec
@@ -12,7 +12,13 @@ URL: https://github.com/docker/scan-cli-plugin/
 Vendor: Docker
 Packager: Docker <support@docker.com>
 
-Requires: docker-ce-cli
+# CentOS 7 and RHEL 7 do not yet support weak dependencies.
+#
+# Note that we're not using <= 7 here, to account for other RPM distros, such
+# as Fedora, which would not have the rhel macro set (so default to 0).
+%if 0%{?rhel} != 7
+Enhances: docker-ce-cli
+%endif
 
 # TODO change once we support scan-plugin on other architectures
 BuildArch: x86_64


rpm/SPECS/docker-scan-plugin.spec
```diff
@@ -12,7 +12,7 @@ URL: https://github.com/docker/scan-cli-plugin/
 Vendor: Docker
 Packager: Docker <support@docker.com>
 
-Requires: docker-ce-cli
+Enhances: docker-ce-cli
```
thaJeztah, line 15: Only thinking if this should be set conditional as well (not sure if older versions just ignore it, or if they would warn/complain), i.e.;

```suggestion
# CentOS 7 and RHEL 7 do not yet support weak dependencies
 %if 0%{?rhel} != 7
 Enhances: docker-ce-cli
 %endif
```

rpm/SPECS/docker-ce-cli.spec
```diff
@@ -18,7 +18,12 @@ Requires: /bin/sh
 Requires: /usr/sbin/groupadd
 # TODO change once we support scan-plugin on other architectures
 %ifarch x86_64
+# CentOS 7 and RHEL 7 do not yet support weak dependencies
+%if 0%{?rhel} == 7
```
thaJeztah, line 22: note that I'm specifically _not_ using `<` or `>`  here, as `rhel` won't be / may not be defined on  other RPM platforms, in which case `0` (default) is picked

crazy-max, line 22:
```suggestion
%if ! (0%{?rhel} && 0%{?rhel} < 7)
```

rpm/SPECS/docker-scan-plugin.spec
thaJeztah: ok; looks like this is needed

```
error: line 15: Unknown tag: Enhances: docker-ce-cli
```

rpm/SPECS/docker-ce-cli.spec
crazy-max: Actually your cond looks right.

thaJeztah: ^^ added a note about that in the comment to prevent someone accidentally changing it

thaJeztah: I only hope this will always be `7` and not `7.1` (e.g.) on CentOS / RHEL 🤔

crazy-max (approved):
LGTM

```
Resolving Dependencies
--> Running transaction check
---> Package docker-ce-cli.x86_64 1:22.04.0-3.el7 will be installed
--> Processing Dependency: docker-scan-plugin(x86-64) for package: 1:docker-ce-cli-22.04.0-3.el7.x86_64
--> Running transaction check
---> Package docker-scan-plugin.x86_64 0:0.17.0-3.el7 will be installed
--> Finished Dependency Resolution

Dependencies Resolved

========================================================================================================================================================================================================== Package                                               Arch                                      Version                                             Repository                                      Size 
==========================================================================================================================================================================================================Installing:
 docker-ce-cli                                         x86_64                                    1:22.04.0-3.el7                                     docker-test                                     18 M 
Installing for dependencies:
 docker-scan-plugin                                    x86_64                                    0.17.0-3.el7                                        docker-test                                    3.7 M 

Transaction Summary
==========================================================================================================================================================================================================Install  1 Package (+1 Dependent package)

Total download size: 22 M
Installed size: 96 M
Is this ok [y/d/N]: y
Downloading packages:
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------Total                                                                                                                                                                     410 MB/s |  22 MB  00:00:00     
Running transaction check
Running transaction test
Transaction test succeeded
Running transaction
  Installing : docker-scan-plugin-0.17.0-3.el7.x86_64                                                                                                                                                 1/2 
  Installing : 1:docker-ce-cli-22.04.0-3.el7.x86_64                                                                                                                                                   2/2 
  Verifying  : 1:docker-ce-cli-22.04.0-3.el7.x86_64                                                                                                                                                   1/2 
  Verifying  : docker-scan-plugin-0.17.0-3.el7.x86_64                                                                                                                                                 2/2 

Installed:
  docker-ce-cli.x86_64 1:22.04.0-3.el7

Dependency Installed:
  docker-scan-plugin.x86_64 0:0.17.0-3.el7

Complete!
```

thaJeztah:
Whoop! Let's get this one in

thaJeztah:
We should probably cherry-pick this one as well

PR #661, opened 2022-03-27
[WIP] mutualize dockerfiles

crazy-max:
experiment with a single dockerfile and multi stages. Also using bake instead of make so we are fully sandboxed.


crazy-max:
Closing, see https://github.com/docker/packaging

PR #667, opened 2022-03-31
[master] Jenkinsfile: add parameters to allow overriding common.mk options

thaJeztah:
This allows overriding some of the options in common.mk, for example
to test a specific branch or PR that has not yet been merged in upstream.


thaJeztah:
🤔 

```
java.lang.IllegalArgumentException: Null value not allowed as an environment variable: GO_VERSION
```

PR #677, opened 2022-04-06
static: make gen-static-ver work natively on macOS

thaJeztah:
macOS uses the BSD flavor of `date`, which does not support the `--date` option
to set a custom time. Previously, we were using an alpine container to provide a
GNU flavor of date, which was a bit of a hack.

This patch rewrites the script to work on macOS directly, without the need of
a container:

    ./static/gen-static-ver . v1.2.3-dev
    0.0.0-20220404154104-b815498

    docker run --rm -v $(pwd):/src -w /src golang bash -c './static/gen-static-ver . v1.2.3-dev'
    0.0.0-20220404154104-b815498

### DIFF
--- a/static/gen-static-ver
+++ b/static/gen-static-ver
@@ -15,11 +15,6 @@ staticVersion="$VERSION"
 if [[ "$VERSION" == *-dev ]]; then
 	export TZ=UTC
 
-	DATE_COMMAND="date"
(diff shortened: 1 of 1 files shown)


crazy-max (approved):

thaJeztah:
Thx!

PR #689, opened 2022-05-12
[master] rpm: use same options for creating symlinks as in the deb builds

thaJeztah:
Found this branch in my local fork, and looks like I didn't open a PR for it yet 😅 

Use the same options for creating the symlinks as for the deb packages, and
add the `-n` and `-f` options;

    -n, --no-dereference        treat LINK_NAME as a normal file if
                                it is a symbolic link to a directory
    -f, --force                 remove existing destination files

### DIFF
--- a/rpm/SPECS/docker-ce-cli.spec
+++ b/rpm/SPECS/docker-ce-cli.spec
@@ -69,7 +69,7 @@ depending on a particular stack or provider.
 %build
 mkdir -p /go/src/github.com/docker
 rm -f /go/src/github.com/docker/cli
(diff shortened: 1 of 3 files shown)


tianon (approved):
It might be worth considering adding `-T` to these as well (avoids the potential for something like `ln -s foo bar` creating `bar/foo` instead of erroring out that `bar` already exists).

thaJeztah:
> It might be worth considering adding `-T` to these as well (avoids the potential for something like `ln -s foo bar` creating `bar/foo` instead of erroring out that `bar` already exists).

Ah, TIL about that option (but I rarely use `ln -s`, other than for the most basic uses) 😂 

Let me add a note to open a PR for adding that option as well 👍

PR #690, opened 2022-05-12
[master] rpmlint: fix "W: no-documentation"

thaJeztah:
These packages were causing a warning that they didn't contain documentation.
For some of these, we need to add man-pages, but at least we can add some of
the files we have. Also adding `AUTHORS` and `SECURITY.md` to existing packages:

    docker-ce.x86_64: W: no-documentation
    docker-ce-rootless-extras.x86_64: W: no-documentation
    docker-compose-plugin.x86_64: W: no-documentation
    docker-scan-plugin.x86_64: W: no-documentation
    The package contains no documentation (README, doc, etc). You have to include
    documentation files.

I'm using unique directory names for each package to collect the docs in, as
we build the packages in the same environment (so possibly they would trip over
each-other), but possibly this is not needed (so "just to be sure").

### DIFF
--- a/rpm/SPECS/docker-ce-cli.spec
+++ b/rpm/SPECS/docker-ce-cli.spec
@@ -99,14 +99,15 @@ install -p -m 644 cli/man/man5/*.5 ${RPM_BUILD_ROOT}%{_mandir}/man5
 install -d ${RPM_BUILD_ROOT}%{_mandir}/man8
 install -p -m 644 cli/man/man8/*.8 ${RPM_BUILD_ROOT}%{_mandir}/man8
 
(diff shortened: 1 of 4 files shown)


thaJeztah:
Looks like this wasn't finished yet (found it locally) 😂 

```
+ for f in AUTHORS LICENSE MAINTAINERS NOTICE README.md
+ install -D -p -m 0644 engine/AUTHORS build-docs/AUTHORS
install: cannot stat 'engine/AUTHORS': No such file or directory
error: Bad exit status from /var/tmp/rpm-tmp.SnUM42 (%install)
```

PR #693, opened 2022-05-13
[master] remove obsolete "selinux, apparmor, seccomp" build-tags from DOCKER_BUILDTAGS

thaJeztah:
### Remove obsolete "selinux" build-tag from DOCKER_BUILDTAGS

Since https://github.com/moby/moby/commit/1c0af18c6c341636ac55644103f922900c616bec (https://github.com/moby/moby/pull/41786), moby no longer uses an selinux build-tag, so we can remove these.

### Remove obsolete "apparmor" build-tag from DOCKER_BUILDTAGS

The apparmor build-tag is no longer used;

```bash
git grep '^// +build.*apparmor'
# (no result)
```

### Remove obsolete "seccomp" build-tag from DOCKER_BUILDTAGS

Starting with https://github.com/moby/moby/commit/c9e19a2aa182e06ebe1653a2c3af02563acf9e36 (https://github.com/moby/moby/pull/42501), the seccomp build-tag is no longer used, so can be removed.


tianon (approved):
:partying_face:

PR #686, opened 2022-05-07
Update "deb" packaging to use "dh_installsystemd"

tianon:
This also removes the now-obsolete upstart script and updates the "debhelper compat level" to 13.

(See also https://lists.debian.org/debian-devel-announce/2020/02/msg00000.html)

Opened as a draft because it's still a WIP -- in theory, this _should_ help with issues like https://github.com/moby/moby/issues/41792 (by triggering `debhelper` to generate the appropriate `systemctl` commands in our `postrm`).

If this is something we want to actually pursue/merge, we probably want to revert the change to `.github/workflows/ci.yml` (which I included just so I could make sure I was successfully building on all the things), and we'll need to update `deb/common/docker-ce.maintscript` to include the correct "next" version number (see https://manpages.debian.org/bullseye/debhelper/dh_installdeb.1.en.html#package~8 and https://manpages.debian.org/bullseye/dpkg/dpkg-maintscript-helper.1.en.html#Removing_a_conffile / https://manpages.debian.org/bullseye/dpkg/dpkg-maintscript-helper.1.en.html#prior-version).

> For example, for a conffile removed in version 2.0-1 of a package, prior-version should be set to 2.0-1~. This will cause the conffile to be removed even if the user rebuilt the previous version 1.0-1 as 1.0-1local1. Or a package switching a path from a symlink (shipped in version 1.0-1) to a directory (shipped in version 2.0-1), but only performing the actual switch in the maintainer scripts in version 3.0-1, should set prior-version to 3.0-1~.

I have not done any testing yet on the actual end result (I was already stretching to squeeze in doing commits periodically to get it to build successfully :see_no_evil:).

cc @thaJeztah

PR #696, opened 2022-05-19
[master] deb: add "Pre-Depends: ${misc:Pre-Depends}" for compat with debhelper > 13

thaJeztah:
Commit f9ac2f67a22f865769a4e74070580276b229ae70 (https://github.com/docker/docker-ce-packaging/pull/686) updated the "debhelper compat
level" to 13, which seems to have introduced a regression when executing the
preRM and postInst scripts;

    invoke-rc.d: syntax error: unknown option "--skip-systemd-native"
       dpkg: error processing package docker-ce (--install):
       installed docker-ce package post-installation script subprocess returned error exit status 1

Searching for the cause of this regression, I found this mention in the debhelper(7)
man-page: https://man7.org/linux/man-pages/man7/debhelper.7.html

    **Retroactively removed in debhelper/13.5:**
    (...)
    This change makes dh_installinit inject a
    misc:Pre-Depends for init-system-helpers (>= 1.54~).
    Please ensure that the package lists
    ${misc:Pre-Depends} in its Pre-Depends field before
    upgrading to compat 12.

This patch adds the recommended Pre-depends

PR #698, opened 2022-05-19
[master] deb: use debhelper-compat=11 for compatibility with ubuntu 18.04

thaJeztah:
Commit f9ac2f67a22f865769a4e74070580276b229ae70 (https://github.com/docker/docker-ce-packaging/pull/686) updated the "debhelper compat
level" to 13 (12?), which seems to have introduced a regression when executing the
preRM and postInst scripts;

    invoke-rc.d: syntax error: unknown option "--skip-systemd-native"
       dpkg: error processing package docker-ce (--install):
       installed docker-ce package post-installation script subprocess returned error exit status 1

This patch tries to downgrade it to 11, which looks to be the default for
ubuntu 18.04 "bionic"


thaJeztah:
Wondering if we need https://github.com/docker/docker-ce-packaging/pull/696 as well (assuming that the "depends" is updated according to the compat-level)

PR #710, opened 2022-06-15
Patch for build with mock

eabdullin1:
Hello!
During the build of Docker, I encountered the fact that it cannot be built using a mock. Changes in the PR allow you to build not from under the user's root, as well as use a mock.


thaJeztah:
I'm not sure if we should make these changes; all the scripts and specs in this repository are currently designed to be used as part of the Dockerfiles in this repository; IIUC, these changes would mean that the distro-provided golang would be installed (not the version as specified in the `GO_IMAGE` build-arg.

PR #665, opened 2022-03-29
fix and refactor static packages

crazy-max:
Signed-off-by: CrazyMax <crazy-max@users.noreply.github.com>

PR #713, opened 2022-06-30
[master] static: remove target_platform and switch to xx

thaJeztah:
- follow-up to https://github.com/docker/docker-ce-packaging/pull/665

I ran into issues running this locally, so moving the last commit to a separate PR


crazy-max:
What is the issue you encountered?

thaJeztah:
@crazy-max initially I thought it was because of a missing SDK, but I had another looks, and it was `xx-info` failing because macOS doesn't have a `/etc/os-release`. I just opened a PR; https://github.com/tonistiigi/xx/pull/72

PR #714, opened 2022-06-30
[master] fix some minor nits

thaJeztah:
### static: create hashes for all files, not just linux

Also switching to alpine, instead of debian:jessy, which is EOL

### Makefile: set initial branch name

Set an initial branch name so that git does not create additional noise
about the default not being configured. Using "master" for now, which
was the previous default; we can probably safely change it to "main",
but leaving that for a future exercise.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ src/github.com/docker/cli:
 	cp -r "$(CLI_DIR)" $@
 else
 src/github.com/docker/cli:
-	git init $@
+	git init --initial-branch=master $@
 	git -C $@ remote add origin "$(DOCKER_CLI_REPO)"
 endif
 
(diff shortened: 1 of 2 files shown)


thaJeztah:
Ah, hm.... looks like we're running on older versions of git;

```
+ make REF=master raspbian-bullseye
Init https://github.com/docker/cli.git
git init --initial-branch=master src/github.com/docker/cli
error: unknown option `initial-branch=master'
usage: git init [-q | --quiet] [--bare] [--template=<template-directory>] [--shared[=<permissions>]] [<directory>]

    --template <template-directory>
                          directory from which templates will be used
    --bare                create a bare repository
    --shared[=<permissions>]
                          specify that the git repository is to be shared amongst several users
    -q, --quiet           be quiet
    --separate-git-dir <gitdir>
                          separate git dir from working tree

Makefile:25: recipe for target 'src/github.com/docker/cli' failed
make: *** [src/github.com/docker/cli] Error 129
script returned exit code 2
```

PR #723, opened 2022-07-31
bump version of Compose to v2.8.0

glours:
Signed-off-by: Guillaume Lours <guillaume.lours@docker.com>


glours:
This version contains breaking changes, we'll skip it

PR #745, opened 2022-08-20
[master] Update GOPROXY to use default with fallback

thaJeztah:
Use the default proxy, to assist with LTS distros that use old git versions but fallback on any error (instead of only on 404 and 410).

From the Go documentation; https://go.dev/ref/mod#goproxy-protocol

> List elements may be separated by commas (,) or pipes (|), which determine error
> fallback behavior. When a URL is followed by a comma, the go command falls back
> to later sources only after a 404 (Not Found) or 410 (Gone) response. When a URL
> is followed by a pipe, the go command falls back to later sources after any error,
> including non-HTTP errors such as timeouts. This error handling behavior lets a
> proxy act as a gatekeeper for unknown modules. For example, a proxy could respond
> with error 403 (Forbidden) for modules not on an approved list (see Private proxy
> serving private modules).

### DIFF
--- a/deb/debian-bullseye/Dockerfile
+++ b/deb/debian-bullseye/Dockerfile
@@ -10,7 +10,7 @@ FROM ${BUILD_IMAGE}
 ARG DEBIAN_FRONTEND=noninteractive
 RUN apt-get update && apt-get install -y curl devscripts equivs git
 
-ENV GOPROXY=direct
+ENV GOPROXY=https://proxy.golang.org|direct
 ENV GO111MODULE=off
 ENV GOPATH /go
 ENV PATH $PATH:/usr/local/go/bin:$GOPATH/bin
--- a/deb/Makefile
+++ b/deb/Makefile
@@ -28,7 +28,6 @@ BUILD?=DOCKER_BUILDKIT=1 \
 RUN_FLAGS=
 RUN?=docker run --rm \
 	-e PLATFORM \
-	-e GOPROXY=https://proxy.golang.org \
 	-e EPOCH='$(EPOCH)' \
 	-e DEB_VERSION=$(word 1, $(GEN_DEB_VER)) \
 	-e VERSION=$(word 2, $(GEN_DEB_VER)) \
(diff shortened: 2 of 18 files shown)


deb/debian-bullseye/Dockerfile
```diff
@@ -10,7 +10,7 @@ FROM ${BUILD_IMAGE}
 ARG DEBIAN_FRONTEND=noninteractive
 RUN apt-get update && apt-get install -y curl devscripts equivs git
 
-ENV GOPROXY=direct
+ENV GOPROXY=https://proxy.golang.org|direct
```
crazy-max, line 13: Is it the right separator?

https://go.dev/ref/mod#environment-variables

> Two keywords may be used in place of proxy URLs:
> 
>     off: disallows downloading modules from any source.
>     direct: download directly from version control repositories instead of using a module proxy.
> 
> GOPROXY defaults to https://proxy.golang.org,direct. Under that configuration, the go command first contacts the Go module mirror run by Google, then falls back to a direct connection if the mirror does not have the module. See https://proxy.golang.org/privacy for the mirror's privacy policy. The GOPRIVATE and GONOPROXY environment variables may be set to prevent specific modules from being downloaded using proxies. See [Privacy](https://go.dev/ref/mod#private-module-privacy) for information on private proxy configuration.

thaJeztah: See the other section on the same page; https://go.dev/ref/mod#goproxy-protocol

> List elements may be separated by commas (,) or pipes (|), which determine error fallback behavior. When a URL is followed by a comma, the go command falls back to later sources only after a 404 (Not Found) or 410 (Gone) response. When a URL is followed by a pipe, the go command falls back to later sources after any error, including non-HTTP errors such as timeouts. This error handling behavior lets a proxy act as a gatekeeper for unknown modules. For example, a proxy could respond with error 403 (Forbidden) for modules not on an approved list (see [Private proxy serving private modules](https://go.dev/ref/mod#private-module-proxy-private)).


See https://github.com/golang/go/issues/37367 and https://github.com/golang/go/commit/69d3a34b176f7e5ac26d289eea66ed5b2dd5f1c8

thaJeztah: I can include that snipped in the commit message (perhaps doesn't hurt)

crazy-max (approved):

thaJeztah:
thx! bringing this one in as well

PR #747, opened 2022-08-20
[20.10 backport] Update GOPROXY to use default with fallback

thaJeztah:
- backport of https://github.com/docker/docker-ce-packaging/pull/745

Use the default proxy, to assist with LTS distros that use old git versions
but fallback on any error (instead of only on 404 and 410).

From the Go documentation; https://go.dev/ref/mod#goproxy-protocol

> List elements may be separated by commas (,) or pipes (|), which determine error
> fallback behavior. When a URL is followed by a comma, the go command falls back
> to later sources only after a 404 (Not Found) or 410 (Gone) response. When a URL
> is followed by a pipe, the go command falls back to later sources after any error,
> including non-HTTP errors such as timeouts. This error handling behavior lets a
> proxy act as a gatekeeper for unknown modules. For example, a proxy could respond
> with error 403 (Forbidden) for modules not on an approved list (see Private proxy
> serving private modules).

(cherry picked from commit 72d51db78ec3a5254bfa94bb201acc2912188ff9)

### DIFF
--- a/deb/Makefile
+++ b/deb/Makefile
@@ -28,7 +28,6 @@ BUILD?=DOCKER_BUILDKIT=1 \
 RUN_FLAGS=
 RUN?=docker run --rm \
 	-e PLATFORM \
-	-e GOPROXY=https://proxy.golang.org \
 	-e EPOCH='$(EPOCH)' \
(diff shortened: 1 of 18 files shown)


crazy-max (approved):

thaJeztah:
all green 🎉

PR #749, opened 2022-08-23
[master] verify: add minimal check for compose, buildx, scan

thaJeztah:
With this patch;

```
+ docker --version
Docker version 0.0.0-20220820144233-1163b46, build 1163b46
++ uname -m
+ '[' x86_64 = x86_64 ']'
+ docker scan --help

Usage:  docker scan [OPTIONS] IMAGE

A tool to scan your images

Options:
      --accept-license    Accept using a third party scanning provider
      --dependency-tree   Show dependency tree with scan results
      --exclude-base      Exclude base image from vulnerability scanning
                          (requires --file)
  -f, --file string       Dockerfile associated with image, provides more
                          detailed results
      --group-issues      Aggregate duplicated vulnerabilities and group
                          them to a single one (requires --json)
      --json              Output results in JSON format
      --login             Authenticate to the scan provider using an
                          optional token (with --token), or web base
                          token if empty
      --reject-license    Reject using a third party scanning provider
      --severity string   Only report vulnerabilities of provided level
                          or higher (low|medium|high)
      --token string      Authentication token to login to the third
                          party scanning provider
      --version           Display version of the scan plugin
+ docker buildx version
github.com/docker/buildx v0.9.1 %{_buildx_gitcommit}
+ docker compose version
Docker Compose version v2.10.0
+ dockerd --version
Docker version 0.0.0-20220820144233-1163b46, build ab37723
+ containerd --version
containerd containerd.io 1.6.7 0197261a30bf81f1ee8e6a4dd2dea0ef95d67ccb
+ runc --version
runc version 1.1.3
commit: v1.1.3-0-g6724737
spec: 1.0.2-dev
go: go1.17.13
libseccomp: 2.5.3
```

PR #750, opened 2022-08-24
[master] buildx: fix missing git-commit in version

thaJeztah:
- relates to https://github.com/docker/docker-ce-packaging/pull/749

### rpm: buildx: fix missing %{_buildx_gitcommit}

The Makefile did not set a `%{_buildx_gitcommit}` variable, causing the version
output to show as:

    docker buildx version
    github.com/docker/buildx v0.9.1 %{_buildx_gitcommit}

### deb: buildx: fix missing git-commit in version

The version output was missing the git-commit, because the deb build
tried to do a ``, but the source files used in the deb builds don't
include the git repository.

Before this:

    + docker buildx version
    github.com/docker/buildx v0.9.1

With this patch

    + docker buildx version
    github.com/docker/buildx v0.9.1 ed00243

### buildx: don't use bash, and enable "trimpath"

Don't use bash to compile buildx, re-format the build script to match
other parts, and enable trimpath.

We could also consider enabling `-s -w` to strip debugging symbols, as
we do for other binaries.

### DIFF
--- a/deb/Makefile
+++ b/deb/Makefile
@@ -9,6 +9,7 @@ GO_IMAGE?=$(GO_BASE_IMAGE):$(GO_VERSION)-buster
 GEN_DEB_VER=$(shell ./gen-deb-ver $(realpath $(CURDIR)/../src/github.com/docker/cli) "$(VERSION)")
 EPOCH?=5
 SCAN_GITCOMMIT?=$(shell cd $(realpath $(CURDIR)/../src/github.com/docker/scan-cli-plugin) && git rev-parse --short HEAD)
+BUILDX_GITCOMMIT?=$(shell cd $(realpath $(CURDIR)/../src/github.com/docker/buildx) && git rev-parse --short HEAD)
 
 ifdef BUILD_IMAGE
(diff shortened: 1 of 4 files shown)


thaJeztah:
Hm... right, so the deb version still isn't right, likely because it tries to use `git` as part of the `deb` build, which likely won't work, as those sources don't contain git;

```
-X github.com/docker/buildx/version.Revision=$(git rev-parse HEAD)
```

```
+ docker buildx version
  github.com/docker/buildx v0.9.1 
```

rpm builds now look correct;

```
+ docker buildx version
github.com/docker/buildx v0.9.1 ed00243
```

thaJeztah:
We should also change Package to `docker-buildx-plugin` or perhaps better `Docker Buildx` instead of the GitHub repository.

crazy-max (approved):
LGTM

In follow-up we could change build command to

```
	&& DESTDIR=/usr/libexec/docker/cli-plugins VERSION=$(BUILDX_VERSION) REVISION=$(BUILDX_GITCOMMIT) make build
```

when https://github.com/docker/buildx/pull/1268 is merged

PR #805, opened 2023-01-02
static: update with latest changes from moby

crazy-max:
follow-up https://github.com/moby/moby/pull/44546
closes #801 

### DIFF
--- a/static/Makefile
+++ b/static/Makefile
@@ -4,6 +4,8 @@ CLI_DIR=$(realpath $(CURDIR)/../src/github.com/docker/cli)
 ENGINE_DIR=$(realpath $(CURDIR)/../src/github.com/docker/docker)
 BUILDX_DIR=$(realpath $(CURDIR)/../src/github.com/docker/buildx)
 
+ENGINE_GITCOMMIT?=$(shell cd $(realpath $(CURDIR)/../src/github.com/docker/docker) && git rev-parse --short HEAD)
+
 GEN_STATIC_VER=$(shell ./gen-static-ver $(CLI_DIR) $(VERSION))
 HASH_CMD=docker run -v $(CURDIR):/sum -w /sum debian:jessie bash hash_files
 DIR_TO_HASH:=build/linux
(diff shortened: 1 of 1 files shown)


static/Makefile
```diff
@@ -9,20 +9,17 @@ HASH_CMD=docker run -v $(CURDIR):/sum -w /sum debian:jessie bash hash_files
 DIR_TO_HASH:=build/linux
 DOCKER_CLI_GOLANG_IMG=golang:$(GO_VERSION)
 
-DOCKER_BUILD_OPTS=
+BAKE_SET_FLAGS=
```
thaJeztah, line 12: The only thing I had with this was that I was wondering if someone may be depending on this variable; wondering if there's a way to keep it for backward-compatibility (might be tricky as it needs a different format for `bake`).

crazy-max: I think we can keep the build args we have previously, let me take a look.

static/Makefile
```diff
@@ -13,18 +15,17 @@ DOCKER_BUILD_OPTS=
 
 ifneq ($(strip $(CONTAINERD_VERSION)),)
 # Set custom build-args to override the containerd version to build for static
-# packages. The Dockerfile for 20.10 and earlier used CONTAINERD_COMMIT, later
-# versions use CONTAINERD_VERSION. We can remove CONTAINERD_VERSION once 20.10.x
-# reaches EOL.
+# packages.
 DOCKER_BUILD_OPTS +=--build-arg=CONTAINERD_VERSION=$(CONTAINERD_VERSION)
-DOCKER_BUILD_OPTS +=--build-arg=CONTAINERD_COMMIT=$(CONTAINERD_VERSION)
```
neersighted, line 20: For other reviews: this comment is out of date; 20.10 now uses `CONTAINERD_VERSION`.

static/Makefile
```diff
@@ -4,6 +4,8 @@ CLI_DIR=$(realpath $(CURDIR)/../src/github.com/docker/cli)
 ENGINE_DIR=$(realpath $(CURDIR)/../src/github.com/docker/docker)
 BUILDX_DIR=$(realpath $(CURDIR)/../src/github.com/docker/buildx)
 
+ENGINE_GITCOMMIT?=$(shell cd $(realpath $(CURDIR)/../src/github.com/docker/docker) && git rev-parse --short HEAD)
```
thaJeztah, line 7: Was wondering; is this different from `ENGINE_REF`? https://github.com/docker/docker-ce-packaging/blob/47167619d0386bee288924b58aad5f71daf1e13f/common.mk#L41

crazy-max: Same as https://github.com/docker/docker-ce-packaging/blob/47167619d0386bee288924b58aad5f71daf1e13f/deb/Makefile#L6

thaJeztah: Ah! Yeah, we should look if we can share more of those (probably the complication was to define the correct path)

thaJeztah (approved):
LGTM, thanks!

neersighted (approved):

PR #810, opened 2023-01-09
rpm: install Linux kernel headers for btrfs on CentOS 7

thaJeztah:
- relates to https://github.com/docker/packaging/pull/69#issuecomment-1374737317
- relates to https://github.com/moby/moby/pull/44761
- relates to https://github.com/moby/moby/pull/44762

### DIFF
--- a/rpm/centos-7/Dockerfile
+++ b/rpm/centos-7/Dockerfile
@@ -23,10 +23,14 @@ RUN if [ -f /etc/yum.repos.d/CentOS-Sources.repo ]; then sed -i 's/altarch/cento
 RUN yum install -y rpm-build rpmlint
 
 # Install Linux kernel headers for btrfs support.
-# See https://github.com/moby/moby/pull/44761 and https://github.com/moby/moby/pull/44762
-RUN rpm --import https://www.elrepo.org/RPM-GPG-KEY-elrepo.org \
- && yum install -y https://www.elrepo.org/elrepo-release-7.el7.elrepo.noarch.rpm \
- && yum --enablerepo=elrepo-kernel install -y kernel-lt-headers
+# See https://github.com/moby/moby/pull/44761 and https://github.com/moby/moby/pull/44762,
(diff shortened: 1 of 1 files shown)


thaJeztah:
Tried to use `sed -i 's/$basearch/x86_64/' /etc/yum.repos.d/elrepo.repo` as a quick hack to make it use x86, but that didn't seem to work either 🤔

thaJeztah:
Looking at one of the mirrors, the package is named `kernel-lt-headers-5.4.227-1.el7.elrepo.x86_64.rpm` (so not a `no arch` 🤔); http://mirrors.colocall.net/elrepo/kernel/el7/x86_64/RPMS/

crazy-max:
> Looking at one of the mirrors, the package is named `kernel-lt-headers-5.4.227-1.el7.elrepo.x86_64.rpm` (so not a `no arch` 🤔); http://mirrors.colocall.net/elrepo/kernel/el7/x86_64/RPMS/

Oh right didn't catch that!

rpm/centos-7/Dockerfile
```diff
@@ -21,6 +21,17 @@ ENV SUITE=${SUITE}
...
+
+# Install Linux kernel headers for btrfs support.
+# See https://github.com/moby/moby/pull/44761 and https://github.com/moby/moby/pull/44762,
+# See https://www.kernel.org/doc/Documentation/kbuild/headers_install.txt
+RUN yum install -y make gcc \
+ && curl -fsSL -O https://github.com/torvalds/linux/archive/refs/tags/v4.12.tar.gz \
+ && tar -xvzf ./v4.12.tar.gz \
+ && rm v4.12.tar.gz \
+ && cd linux-4.12/ \
+ && make headers_install ARCH=arm64 INSTALL_HDR_PATH=/usr
```
neersighted, line 33: It seems like we shouldn't be setting ARCH here, should we? I don't quite see how this Dockerfile is ARM-specific.

thaJeztah:
We'll probably go for https://github.com/docker/docker-ce-packaging/pull/811 instead

PR #811, opened 2023-01-09
rpm: disable btrfs on CentOS 7 / RHEL 7 as it's deprecated

thaJeztah:
The `btrfs` storage driver on CentOS and RHEL was provided as a technology preview by CentOS and RHEL, but has been deprecated since the Red Hat Enterprise Linux 7.4 release:

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/storage_administration_guide/ch-btrfs

This patch disables the driver for CentOS 7 (it was already disabled for CentOS 8 and up).


crazy-max (approved):

thaJeztah:
We discussed this in the maintainers call, and we're okay with just disabling it; let me bring this one in

PR #821, opened 2023-01-11
update example versions

thaJeztah:
Starting with Docker 23.0.0, we're moving away from CalVer, instead using SemVer(ish) versions. This aptch updates some examples used to match the new format.


thaJeztah:
@crazy-max @vvoland ptal 😅

crazy-max (approved):

PR #830, opened 2023-01-12
[master] remove libbtrfs-dev, btrfs-tools, btrfs-progs-devel build-dependencies

thaJeztah:
- relates to https://github.com/moby/moby/pull/44761
- relates to https://github.com/moby/moby/pull/44707
- relates to https://github.com/moby/moby/issues/44698
- relates to https://github.com/moby/moby/pull/44762

Starting with https://github.com/moby/moby/commit/6834304febca3c0610d1061bd9612ee99020df34, building the engine no longer requires these headers, as it now uses the kernel UAPI, which means we no longer need these packages as dependency.

Note that the static builds may still need them for containerd, but dependencies for static builds are defined in the Dockerfile (containerd is also in the process of migrating to use the kernel UAPI).


thaJeztah:
@crazy-max @neersighted ptal 🤗

neersighted (approved):

crazy-max (approved):

thaJeztah:
Thx!

PR #718, opened 2022-07-06
[master] fix and refactor static packages with fixed cross-comp on moby

crazy-max:
follow-up https://github.com/docker/docker-ce-packaging/pull/665#issuecomment-1173088399

this builds the docker engine from https://github.com/moby/moby/pull/43529 which fixes issues with cross compilation to remove unneeded in the current logic.

it also fixes cross comp for extra tools (containerd, runc, tini, etc...) which is currently not cover on master branch but fixed in https://github.com/moby/moby/pull/43529.

see last commit

PR #835, opened 2023-01-25
refactor static packages

crazy-max:
closes #718
closes #665

This is a follow-up #718 and #665. Now that docker-engine supports cross-compilation with `--platform` https://github.com/moby/moby/pull/44546 and compose too https://github.com/docker/compose/pull/9744 we can move on better support for static packages and mutualize job steps and build logic.

Versioning for plugins doesn't go through the `gen-static-ver` script. Let me know if we should.


Jenkinsfile
```diff
@@ -54,78 +54,57 @@ def genBuildStep(LinkedHashMap pkg, String arch) {
...
+    return opts.arches.collectEntries {
+        ["${opts.image}-${it}": genPkgStep(opts, it)]
+    }
+}
+
+def genStaticStep(LinkedHashMap pkg, String arch) {
+    def config = [
+        amd64:   [label: "x86_64",  targetarch: "amd64"],
+        aarch64: [label: "aarch64", targetarch: "arm64"],
+        armv6:   [label: "aarch64", targetarch: "arm/v6"],
```
crazy-max, line 67: I think `armv6` is fine for `armel` builds. Don't think we want to support `armv5` anymore?

crazy-max:
@thaJeztah As discussed added extra commit to keep the same file format for bundles so for now we keep using current arches to reflect folder structure on [download.docker.com](https://download.docker.com/linux/static/stable/)

Also adds another commit that creates `docker.tgz` archive for backward compat.

crazy-max:
Will open a follow-up on release-repo and release-packaging to be sure everything works fine.

Jenkinsfile
tianon: I'm 100% fine with us not supporting `armv5` anymore, but I do not think we should call anything _else_ `armel` :sweat_smile:

crazy-max: > but I do not think we should call anything _else_ `armel` 😅

Totally agree, we need a better folder structure in the future to avoid this kind of confusion

PR #838, opened 2023-01-30
deb: add Debian / Raspbian 12 "bookworm" (next stable)

thaJeztah:
Not yet released, but freeze was announced;
https://lists.debian.org/debian-devel-announce/2023/01/msg00004.html


tianon (approved):

tianon:
(To elaborate a little -- this is only the _first_ stage of freeze, but it does mean that the target should be reasonably stable at least in the toolchain and thus not cause too many problems for us to start publishing for now.)

thaJeztah:
Yes; so the overall idea was to at least have it running in CI on this repo; and potentially "soft release" packages for users to start testing it.

thaJeztah:
Let me bring this one in; thanks!

PR #874, opened 2023-03-31
[23.0] remove scan-cli-plugin as recommends/required dependency

thaJeztah:
The scan-cli-plugin has been deprecated, and will no longer receive updates, so we don't want it to be automatically installed as a dependency; https://github.com/docker/scan-cli-plugin/blob/b69ac460a6c23243b858326f0e26f6bd202ca781/internal/deprecation.go#L13-L15


bsousaa (approved):

thaJeztah:
Thx!

PR #901, opened 2023-05-23
[23.0 backport] update buildx to v0.10.5

crazy-max:
* backport of #900


thaJeztah:
Thinking of this; we can _probably_ stop building the plugins from this branch (as they'll be built as part of the 24.0.x release now) (but we should probably start a 24.0 branch, as we need to remove the device mapper stuff for 25.0)

I'll keep both for a later exercise 😅

PR #910, opened 2023-06-21
[23.0 backport] deb, rpm: update golang image to use bookworm instead of buster

thaJeztah:
- backport of https://github.com/docker/docker-ce-packaging/pull/909

While we only use the golang binaries from these images (which should be the same for both), Debian buster reached EOL. Update the Golang image to the current stable version of Debian (Debian 12 "bookworm")

(cherry picked from commit 46f56f2720c6f36650294e728079e21609353702)


thaJeztah:
@neersighted @crazy-max PTAL

crazy-max (approved):

neersighted (approved):

Issue #899, opened 2023-05-22
Use upstream init scripts


neersighted:
Per https://github.com/moby/moby/pull/45548, we should start pulling in the upstream init scripts instead of maintaining parallel copies in this repo.

Issue #851, opened 2023-02-23
Uninstalling docker-ce RPM does not disable docker.socket


ruizink:
The `docker-ce` rpm package does not cleanup the `docker.socket` unit from systemd on uninstall.

This makes it so that `docker.socket` keeps in an `active` state, despite the socket not being functional:

```
# systemctl status docker.socket
● docker.socket
   Loaded: not-found (Reason: Unit docker.socket not found.)
   Active: active (listening) since Thu 2023-02-23 22:57:22 UTC; 33s ago
    Tasks: 0 (limit: 25092)
   Memory: 264.0K
   CGroup: /system.slice/docker.socket

Feb 23 22:57:22 751cf9ad8325 systemd[1]: Starting Docker Socket for the API.
Feb 23 22:57:22 751cf9ad8325 systemd[1]: Listening on Docker Socket for the API.
Feb 23 22:57:53 751cf9ad8325 systemd[1]: docker.socket: Socket unit configuration has changed while unit has been running, no open socket file descriptor left. The socket unit is not functional until restarted.
```

This is a silent trap for the next time we start the service after reinstallation, because socket activation won't work:

```
# journalctl -u docker.service

Feb 23 23:03:37 751cf9ad8325 systemd[1]: Starting Docker Application Container Engine...
Feb 23 23:03:38 751cf9ad8325 dockerd[19450]: time="2023-02-23T23:03:38.028768000Z" level=info msg="Starting up"
Feb 23 23:03:38 751cf9ad8325 dockerd[19450]: failed to load listeners: no sockets found via socket activation: make sure the service was started by systemd
Feb 23 23:03:38 751cf9ad8325 systemd[1]: docker.service: Main process exited, code=exited, status=1/FAILURE
Feb 23 23:03:38 751cf9ad8325 systemd[1]: docker.service: Failed with result 'exit-code'.
Feb 23 23:03:38 751cf9ad8325 systemd[1]: Failed to start Docker Application Container Engine.
Feb 23 23:03:40 751cf9ad8325 systemd[1]: docker.service: Service RestartSec=2s expired, scheduling restart.
Feb 23 23:03:40 751cf9ad8325 systemd[1]: docker.service: Scheduled restart job, restart counter is at 1.
Feb 23 23:03:40 751cf9ad8325 systemd[1]: Stopped Docker Application Container Engine.
```

corhere:
The deb package is also affected: https://github.com/moby/moby/issues/45059

[...]

PR #699, opened 2022-05-19
[master] Revert "Update "deb" packaging to use "dh_installsystemd""

thaJeztah:
This reverts commit f9ac2f67a22f865769a4e74070580276b229ae70 (https://github.com/docker/docker-ce-packaging/pull/686), as it unfortunately won't work with ubuntu 18.04

We can work on this later, to see if we can find a suitable workaround

PR #914, opened 2023-06-21
[master] Update "deb" packaging to use "dh_installsystemd", and remove upstart script

thaJeztah:
- depends on https://github.com/docker/docker-ce-packaging/pull/911
- reverts https://github.com/docker/docker-ce-packaging/pull/699
- re-applies https://github.com/docker/docker-ce-packaging/pull/686
- closes https://github.com/docker/docker-ce-packaging/pull/698
- closes https://github.com/docker/docker-ce-packaging/pull/696
- partially addresses https://github.com/docker/docker-ce-packaging/issues/899
- relates to https://github.com/moby/moby/pull/45548
- fixes / addresses https://github.com/moby/moby/issues/44901
- fixes / addresses https://github.com/docker/docker-ce-packaging/issues/851

This reverts commit cafc21c4f63c08491ace9ba702785493bae73b8a.

This also removes the now-obsolete upstart script and updates the "debhelper compat level" to 12 (which necessitates adding "override_dh_dwz" -- safe to remove once we can update to debhelper-compat 13+).

(See also https://lists.debian.org/debian-devel-announce/2020/02/msg00000.html)


deb/common/docker-ce.docker.service
```diff
@@ -0,0 +1 @@
+engine/contrib/init/systemd/docker.service
```
tianon, line 1: This symlink ends up in `debian/` so needs to be `../` prefixed, right? :eyes:

```suggestion
../engine/contrib/init/systemd/docker.service
```

deb/common/docker-ce.docker.socket
```diff
@@ -0,0 +1 @@
+engine/contrib/init/systemd/docker.socket
```
tianon, line 1: ditto:

```suggestion
../engine/contrib/init/systemd/docker.socket
```

deb/common/docker-ce.maintscript
```diff
@@ -0,0 +1,3 @@
+# *technically* this "5:21~" should be something more specific like "5:22.06.1-1~" (the first released version to introduce this "rm_conffile" line), but as long as it's strictly *newer* than the last version to ship the conffile and stricly *older* than the first version with the rm_conffile line, it should work fine (so we're using the will-never-exist major version of "21" as our in-between release for the 20.xx to 22.xx)
+rm_conffile /etc/init/docker.conf 5:21~
+# this should stick around in the package until there are no more users who will ever upgrade from "5:20.xx" to something newer (it's reasonably harmless for this to stay indefinitely)
```
tianon, line 3: I recognize my prose!!  You can't fool me!! :joy: :heart:

However, this should be referencing the version it's going to be _included_ in, so here, this should be v25 (`5:25~`, to be precise), but in backports to older releases this can be more specifically the version number of the backport (for example, `5:24.0.3~` if we put this into 24.0.3).

```suggestion
# *technically* this "5:25~" should be something more specific like "5:25.0.0-1~" (the first released version to introduce this "rm_conffile" line), but as long as it's strictly *newer* than the last version to ship the conffile and stricly *older* than the first version with the rm_conffile line, it should work fine (so we're using the will-never-exist non-semver version of "25" as our in-between release for the 24.x.x to 25.x.x)
rm_conffile /etc/init/docker.conf 5:25~
# this should stick around in the package until there are no more users who will ever upgrade from "5:24.x.x" to something newer (it's reasonably harmless for this to stay indefinitely)
```

tianon (commented):
Generally I like the shape, just a few suggestions :heart:

deb/common/docker-ce.docker.service
tianon: Curious that the CI builds succeeded; I bet they're missing systemd service files :grimacing:

neersighted (changes requested):
LGTM with @tianon's tweaks

deb/common/docker-ce.docker.service
thaJeztah: oh, interesting; so this was a straight revert of your original one; wondering if things were broken there, or is it because we moved things?

tianon: Oh, it was probably broken there!! :facepalm:

thaJeztah: I haven't checked yet 🙈 

I'm always confused with these paths as well, because it's possible we put the directories in a different path during build (which I need to check)

tianon: Yeah, all these files end up under `debian/`, of which `engine/` is a "sibling directory" :sweat_smile:

neersighted:
Applied the review comments; @tianon @thaJeztah PTAL

deb/common/docker-ce.docker.service
```diff
@@ -0,0 +1 @@
+../engine/contib/init/systemd/docker.service
```
tianon, line 1:
```suggestion
../engine/contrib/init/systemd/docker.service
```

deb/common/docker-ce.docker.socket
```diff
@@ -0,0 +1 @@
+../engine/contib/init/systemd/docker.socket
```
tianon, line 1:
```suggestion
../engine/contrib/init/systemd/docker.socket
```

deb/common/docker-ce.maintscript
```diff
@@ -0,0 +1,3 @@
+# *technically* this "5:24~" should be something more specific like "5:24.0.5-1~" (the first released version to introduce this "rm_conffile" line), but as long as it's strictly *newer* than the last version to ship the conffile and stricly *older* than the first version with the rm_conffile line, it should work fine (so we're using the will-never-exist non-semver version of "24" as our in-between release for the 23.x.x to 24.x.x)
+rm_conffile /etc/init/docker.conf 5:24~
```
tianon, line 2: This should actually be `5:25~` for 25+ and a much more specific minor version for 24.x backport, right?

tianon (commented):
very close :see_no_evil: :heart:

deb/common/docker-ce.docker.service
neersighted: Oh my god, it's been a long day.

deb/common/docker-ce.maintscript
neersighted: I was intending to apply this commit to 24.0 as we haven't branched yet; it sounds like we should go with the patch version here, and only use the major version in the 25.0 branch?

tianon: Yeah, fair - so this should either be `5:24.0.5~` (which would honestly be fine for both 24.x backport and 25.x+ if you want the same value in both and we actually get this into 24.0.5) or you could `5:25~` for 25.x+.

tianon: (The problem is that as-is, a bare "24" is _not_ newer than 24.0.5, so doesn't satisfy my long comments :smile:)

neersighted: Check the last push, it's currently `rm_conffile /etc/init/docker.conf 5:24.0.5-1~`

tianon (approved):
Technically the `-1` is unnecessarily specific, but should be harmless.  LGTM :+1: :heart:

PR #946, opened 2023-09-21
[master] bump compose version to v2.22.0

glours:
https://github.com/docker/compose/releases/tag/v2.22.0

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -38,7 +38,7 @@ DOCKER_BUILDX_REPO  ?= https://github.com/docker/buildx.git
 REF                ?= HEAD
 DOCKER_CLI_REF     ?= $(REF)
 DOCKER_ENGINE_REF  ?= $(REF)
-DOCKER_COMPOSE_REF ?= v2.21.0
+DOCKER_COMPOSE_REF ?= v2.22.0
 DOCKER_BUILDX_REF  ?= v0.11.2
 
 # Use "stage" to install dependencies from download-stage.docker.com during the


common.mk
```diff
@@ -17,7 +17,7 @@ CHOWN:=docker run --rm -v $(CURDIR):/v -w /v alpine chown
 DEFAULT_PRODUCT_LICENSE:=Community Engine
 PACKAGER_NAME?=
 DOCKER_GITCOMMIT:=abcdefg
-GO_VERSION:=1.20.8
+GO_VERSION:=1.21.0
```
neersighted, line 20: Can we split this into a separate commit (so "bump Go" then "bump Compose")?

glours: Sure

glours: reverted

sam-thibault (approved):

neersighted:
Copying from Slack:

We have legit CI failures here as Compose now requires Go 1.21; this is blocked until either:
* We push a version that uses a separate Go compiler for Compose (I'd be okay with this)
* All other projects (e.g. CLI, Engine, buildx) compile with Go 1.21

Since the second option will likely happen soon (read: days to weeks), @glours expressed he's fine with waiting to get this in.

deb/common/rules
```diff
@@ -33,7 +33,7 @@ override_dh_auto_build:
 
 	# Build the compose plugin
 	cd /go/src/github.com/docker/compose \
-	&& make VERSION=$(COMPOSE_VERSION) DESTDIR=/usr/libexec/docker/cli-plugins build
+	&& make GO_VERSION=$(GO_VERSION_COMPOSE) VERSION=$(COMPOSE_VERSION) DESTDIR=/usr/libexec/docker/cli-plugins build
```
thaJeztah, line 36: This probably won't work, because both `deb` and `rpm` build-tools don't / can't use docker; they're plain `make` targets (but may be running inside a docker container)

glours: Yep I saw that, I'll test with the dedicated bake command defined in Compose makefile and check if using a container build will do the work

thaJeztah:
oh, LOL, I did a rebase earlier in ca you still wanted to look at a custom option; https://github.com/docker/docker-ce-packaging/pull/953

Let's merge this one

thaJeztah (approved):
LGTM

PR #958, opened 2023-10-18
bump compose version to v2.23.0

glours:
Upgrade version of Compose to `v2.23.0`


neersighted (approved):
Will need a backport to 24.0 as well.

glours:
@neersighted it won't compile as the 24 branch is still using golang 1.20, no?

thaJeztah:
Yup, backport won't work; we can't publish new packages of compose until we do the v25.0.0 release

PR #961, opened 2023-11-06
deb, rpm: set minimum containerd version to 1.6.24

thaJeztah:
relates to https://github.com/moby/moby/pull/46778

### DIFF
--- a/deb/common/control
+++ b/deb/common/control
@@ -25,7 +25,7 @@ Vcs-Git: git://github.com/docker/docker.git
 
 Package: docker-ce
 Architecture: linux-any
-Depends: containerd.io (>= 1.6.4),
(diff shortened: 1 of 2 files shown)


thaJeztah:
@neersighted @cpuguy83 @corhere ptal

neersighted (approved):

corhere (commented):
Is there a way to express that containerd 1.7 < v1.7.3 are incompatible?

thaJeztah:
Maybe there is, but our package repos have never shipped packages for containerd 1.7, so if we would publish those, we would start with 1.7.8 or higher

corhere (approved):
Oh right, the major distros call their packages `containerd`, which would not satisfy a dependency on `containerd.io`

PR #968, opened 2023-11-21
rpm: patch RLIMIT_NOFILE for EL 7 systems

neersighted:
RLIMIT_NOFILE defaults to an unusable value on EL 7 systems, but more modern systems have sane defaults. As such, patch the systemd unit only on EL 7 systems to ensure correct behaviour.

### DIFF
--- a/rpm/patches/01-rlimit_nofile-rhel7.patch
+++ b/rpm/patches/01-rlimit_nofile-rhel7.patch
@@ -1,0 +1,15 @@
+diff --git i/engine/contrib/init/systemd/docker.service w/engine/contrib/init/systemd/docker.service
+index d8c7867057..b73ecf7363 100644
+--- i/engine/contrib/init/systemd/docker.service
++++ w/engine/contrib/init/systemd/docker.service
+@@ -30,6 +30,10 @@ StartLimitInterval=60s
+ # in the kernel. We recommend using cgroups to do container-local accounting.
+ LimitNPROC=infinity
+ LimitCORE=infinity
++# Older systemd versions default to a LimitNOFILE of 1024:1024, which is insufficient for many
++# applications including dockerd itself and will be inherited. Raise the hard limit, while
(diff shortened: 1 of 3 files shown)


thaJeztah:
Looks like containerd did not (yet) backport that change to 1.6 and 1.7, but we need to keep an eye in case they do;

- https://github.com/containerd/containerd/blob/v1.7.9/containerd.service
- https://github.com/containerd/containerd/blob/v1.6.25/containerd.service

Or we could consider doing the reverse, and patching the containerd systemd unit to (conditionally) remove the `LimitNOFILE=infinity`

rpm/patches/01-rlimit_nofile-rhel7.patch
```diff
@@ -0,0 +1,15 @@
...
+ # in the kernel. We recommend using cgroups to do container-local accounting.
+ LimitNPROC=infinity
+ LimitCORE=infinity
++# Older systemd versions default to a LimitNOFILE of 1024:1024, which is insufficient for many
++# applications including dockerd itself and will be inherited. Raise the hard limit, while
++# preserving the soft limit for select(2).
++LimitNOFILE=1024:524288
+ 
+ # Comment TasksMax if your systemd version does not support it.
+ # Only systemd 226 and above support this option.
```
thaJeztah, line 15: Unrelated, but now wondering what version of systemd RHEL/CentOS 7 are running, and if this comment is still relevant (ISTR it was always just a warning, so not even sure if relevant _at all_)

neersighted: 219: https://centos.pkgs.org/7/centos-x86_64/systemd-sysv-219-78.el7.x86_64.rpm.html

neersighted:
Yeah, thought I had this working but I was disentangling it from other changes I was testing last week; will fiddle with the paths a bit.

thaJeztah (approved):
LGTM

PR #986, opened 2024-02-02
remove Ubuntu Lunar 23.04, as it's EOL

thaJeztah:
Ubuntu 23.04 reached EOL on January 25, 2024, so should no longer be used. https://wiki.ubuntu.com/Releases#End_of_Life


krissetto (approved):
LGTM

PR #991, opened 2024-02-22
RHEL PoC

neersighted:
This is a PoC for building RHEL packages assuming that a RHEL username/password is available in the environment.

This is mostly intended for internal use by Docker, as it is assumed that the main utility is performing builds with a 'real RHEL' for enterprise customers.

However, when a unencumbered RHEL-equivalent distribution (e.g. Rocky) is supported, it will make sense to generalize the 'rhel' infrastructure in this repo.

### DIFF
--- a/rpm/rhel-8/Dockerfile
+++ b/rpm/rhel-8/Dockerfile
@@ -1,0 +1,36 @@
+# syntax=docker/dockerfile:1
+
+ARG GO_IMAGE
+ARG DISTRO=rhel
+ARG SUITE=8
+ARG BUILD_IMAGE=registry.access.redhat.com/ubi8/ubi
+
+FROM ${GO_IMAGE} AS golang
+
+FROM ${BUILD_IMAGE} AS subscribed-image
+ARG RH_USER
+ARG RH_PASS
+RUN rm /etc/rhsm-host 
+RUN subscription-manager register --username=$RH_USER --password=$RH_PASS
+RUN subscription-manager repos --enable codeready-builder-for-rhel-8-$(arch)-rpms
+# RUN dnf config-manager --set-enabled codeready-builder-for-rhel-8-$(arch)-rpms
+
+FROM subscribed-image
+
+ENV GOPROXY=https://proxy.golang.org|direct
+ENV GO111MODULE=off
+ENV GOPATH=/go
+ENV PATH $PATH:/usr/local/go/bin:$GOPATH/bin
+ENV AUTO_GOPATH 1
+ENV DOCKER_BUILDTAGS exclude_graphdriver_btrfs
+ARG DISTRO
+ARG SUITE
+ENV DISTRO=${DISTRO}
+ENV SUITE=${SUITE}
+
+RUN dnf install -y rpm-build rpmlint
+COPY --link SPECS /root/rpmbuild/SPECS
+RUN dnf builddep -y /root/rpmbuild/SPECS/*.spec
+COPY --link --from=golang /usr/local/go /usr/local/go
+WORKDIR /root/rpmbuild
+ENTRYPOINT ["/bin/rpmbuild"]
--- a/rpm/rhel-8/Dockerfile
+++ b/rpm/rhel-8/Dockerfile
@@ -8,12 +8,20 @@ ARG BUILD_IMAGE=registry.access.redhat.com/ubi8/ubi
 FROM ${GO_IMAGE} AS golang
 
 FROM ${BUILD_IMAGE} AS subscribed-image
-ARG RH_USER
-ARG RH_PASS
-RUN rm /etc/rhsm-host 
-RUN subscription-manager register --username=$RH_USER --password=$RH_PASS
-RUN subscription-manager repos --enable codeready-builder-for-rhel-8-$(arch)-rpms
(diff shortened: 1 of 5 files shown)


rpm/rhel-7/Dockerfile
```diff
@@ -1,30 +0,0 @@
-# syntax=docker/dockerfile:1
```
thaJeztah, line 1: This one may be in use by IBM?

rpm/rhel-8/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+ARG BUILD_IMAGE=registry.access.redhat.com/ubi8/ubi
+
+FROM ${GO_IMAGE} AS golang
+
+FROM ${BUILD_IMAGE} AS subscribed-image
+ARG RH_USER
+ARG RH_PASS
+RUN rm /etc/rhsm-host 
+RUN subscription-manager register --username=$RH_USER --password=$RH_PASS
+RUN subscription-manager repos --enable codeready-builder-for-rhel-8-$(arch)-rpms
```
thaJeztah, line 15: Didn't this produce a subscription-file, so something we could run in CI, and  `--mount` ? Passing these as argument means they leak into the image produced (and build-cache).

neersighted: This is somewhat intentional for now; the cleanup step is "manual" (read: a cron job elsewhere) as a stopgap while I work on lifecycle robustness. That being said, I hadn't considered build cache issues since we don't actually clean the cache on the Jenkins nodes, as best I can tell...

Given that, maybe it would be better to use fixed certificates; this is somewhat messier (mostly for containerd-packaging) and introduces the manual update process/point of failure, but I can mock it up as an alternative (and it does obviate the runtime lifecycle issues).

rpm/rhel-8/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+ARG SUITE=8
+ARG BUILD_IMAGE=registry.access.redhat.com/ubi8/ubi
+
+FROM ${GO_IMAGE} AS golang
+
+FROM ${BUILD_IMAGE} AS subscribed-image
+ARG RH_USER
+ARG RH_PASS
+RUN rm /etc/rhsm-host 
+RUN subscription-manager register --username=$RH_USER --password=$RH_PASS
```
AkihiroSuda, line 14: This should use `RUN --mount=type=secret` to avoid leaking the password in the image history (although the image is never pushed to public)

vvoland:
Rebased

thaJeztah:
rebased

thaJeztah:
Rebased

thaJeztah:
rebased after https://github.com/docker/docker-ce-packaging/pull/1030 was merged /cc @vvoland

vvoland:
Rebased after https://github.com/docker/docker-ce-packaging/pull/1031

vvoland:
Rebased after https://github.com/docker/docker-ce-packaging/pull/1032

vvoland:
Rebased after https://github.com/docker/docker-ce-packaging/pull/1034

vvoland:
Rebased after https://github.com/docker/docker-ce-packaging/pull/1038

thaJeztah:
Rebased after https://github.com/docker/docker-ce-packaging/pull/1039

thaJeztah:
rebased after https://github.com/docker/docker-ce-packaging/pull/1041

thaJeztah:
rebased after https://github.com/docker/docker-ce-packaging/pull/1043

vvoland:
Rebased after #1050

thaJeztah:
@vvoland I pushed a commit to change the `--build-arg` to use a build-secret instead; Taking a similar approach as is used for containerd-packaging; also ran a test-build in the release-packaging pipeline, and it looks to be happy.

thaJeztah:
If the change LGTY, I can squash the last two commits

vvoland (approved):
LGTM; also fine to leave the commits separate IMO

rpm/rhel-9/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+	rm -f /etc/rhsm-host
+
+	if [ ! -f /run/secrets/rh-user ] || [ ! -f /run/secrets/rh-pass ]; then
+		echo "Either RH_USER or RH_PASS is not set. Running build without subscription."
+	else
+		subscription-manager register \
+			--username="$(cat /run/secrets/rh-user)" \
+			--password="$(cat /run/secrets/rh-pass)"
+
+		subscription-manager repos --enable codeready-builder-for-rhel-9-$(arch)-rpms
```
thaJeztah, line 21: FWIW; still looking at the `$(arch)` bit here; is this expanding using a RPM arg or is this a bug, just “happened to work”?

thaJeztah: DOH! Of course; was with my head in env-vars and build-args, and I was thinking it was depending on some env-var or RPM macro/var 🙈

thaJeztah:
>  also fine to leave the commits separate IMO

Yeah, perhaps keeping the commit separate is still fine.

thaJeztah:
Let me bring this one in 👍

PR #1001, opened 2024-03-14
debian: add explicit GOARM for raspbian

neersighted:
This is a quick band-aid; we need to overhaul how we handle architectures in these scripts (and use the Debian base images for Raspbian); but for now this should ensure we force ARMv6 for Raspbian targets.


thaJeztah (approved):
LGTM

PR #1003, opened 2024-03-15
debian: build armv6 binaries for armhf

neersighted:
Debian armhf is armv7 and hard-float, and Raspbian is a rebuild of Debian with armv6 and hard-float. This is done to make use of the BCM2835 FPU, as while the Debian armel (armv5 and soft-float) port works, it does not make full use of the CPU's hardware.

By making this change, our 'armhf' binaries will work on both armv6 and armv7 systems.

### DIFF
--- a/deb/common/rules
+++ b/deb/common/rules
@@ -1,17 +1,19 @@
 #!/usr/bin/make -f
 
+# Include default Makefile variables.
+include /usr/share/dpkg/default.mk
+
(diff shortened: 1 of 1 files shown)


tianon (approved):

tianon-sso (approved):

thaJeztah (dismissed):
LGTM

thaJeztah (approved):
LGTM

PR #554, opened 2021-06-29
Fix build-script for docker-proxy (after libnetwork integration) and update to Go 1.16.5

thaJeztah:
libnetwork has been integrated in the docker/docker (moby) repository, and now is compiled as part of hack/make.sh dynbinary.

relates to https://github.com/moby/moby/pull/42262 and https://github.com/moby/moby/pull/42539

PR #314, opened 2019-03-25
[master] [TAR-610] [TAR-611] Build the daemon as part of the package building process

seemethere:
Basically a reverse of #223.

Should hopefully resolve issues:
* https://github.com/moby/moby/issues/38175
* https://github.com/docker/for-linux/issues/452

Tested with `docker engine activate` and `docker engine update`, which work as expected.

cc @andrewhsu @thaJeztah

PR #1010, opened 2024-04-02
rpm: drop unnecessary build-deps

neersighted:
**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/rpm/SPECS/docker-ce.spec
+++ b/rpm/SPECS/docker-ce.spec
@@ -41,7 +41,6 @@ BuildRequires: pkgconfig
 BuildRequires: pkgconfig(systemd)
 BuildRequires: systemd-devel
 BuildRequires: tar
-BuildRequires: which
 
 # conflicting packages
 Conflicts: docker
--- a/rpm/SPECS/docker-ce.spec
+++ b/rpm/SPECS/docker-ce.spec
@@ -34,7 +34,6 @@ BuildRequires: cmake
 BuildRequires: gcc
 BuildRequires: git
 BuildRequires: glibc-static
-BuildRequires: libarchive
 BuildRequires: libtool
 BuildRequires: libtool-ltdl-devel
 BuildRequires: make


thaJeztah:
Also curious if we need to look at updating the `Standards-Version`, and possibly if there's improvements we can use from newer versions; https://www.debian.org/doc/debian-policy/upgrading-checklist.html#version-3-9-6

At least I spotted some "-Arch" options (perhaps useful if we need architecture specific handling), and saw some mentions of fallbacks that are no longer needed.

https://github.com/docker/docker-ce-packaging/blob/1e9d8a686ed9e293ca8c0017fb1242f07ab6e691/deb/common/control#L21

rpm/SPECS/docker-ce.spec
```diff
@@ -34,15 +34,13 @@ BuildRequires: cmake
...
 BuildRequires: glibc-static
-BuildRequires: libarchive
 BuildRequires: libtool
 BuildRequires: libtool-ltdl-devel
 BuildRequires: make
 BuildRequires: pkgconfig
 BuildRequires: pkgconfig(systemd)
 BuildRequires: systemd-devel
 BuildRequires: tar
-BuildRequires: which
```
thaJeztah, line 45: This one was originally added in https://github.com/docker/docker-ce-packaging/commit/fda74c4df83554ae2662184b4a5a1a1eaaf6927a (through https://github.com/docker/docker-ce-packaging/pull/314).


My _guess_ is that some script at some point incorrectly used `which <some command>` instead of `command -v <some command>`, but I couldn't find a reference to it either at the commit that added it, nor at current master, so I'm gonna assume here that it was never needed at all;


```bash
git rev-parse --verify HEAD
fda74c4df83554ae2662184b4a5a1a1eaaf6927a

git grep 'which '
deb/common/docker-ce.docker.init:       # see also init_is_upstart in /lib/lsb/init-functions (which isn't available in Ubuntu 12.04, or we'd use it directly)
deb/gen-deb-ver:    # be compared to determine which happened later, the commit hash identifes
rpm/gen-rpm-ver:    # be compared to determine which happened later, the commit hash identifes
static/gen-static-ver:    # be compared to determine which happened later, the commit hash identifes

git grep 'command '
deb/common/docker-ce.postinst:          echo "You can use the 'docker engine update' command to update your system, or switch to using the Enterprise packages."
deb/common/rules:       dh $@ --with=bash-completion $(shell command -v dh_systemd_enable > /dev/null 2>&1 && echo --with=systemd)
deb/gen-deb-ver:    # as a pre-release before version v0.0.0, so that the go command prefers any
rpm/gen-rpm-ver:    # as a pre-release before version v0.0.0, so that the go command prefers any
static/gen-static-ver:    # as a pre-release before version v0.0.0, so that the go command prefers any


git checkout master

git rev-parse --verify HEAD
2f0c865a12578226abe568ca772c9ca06e58516b

git grep 'which '
Jenkinsfile:        // Otherwise it accidentally pulls armel images which then breaks the verify step
Jenkinsfile:                // in Jenkins' BlueOcean view, which truncates names....
Jenkinsfile:                // in Jenkins' BlueOcean view, which truncates names....
Jenkinsfile:                // in Jenkins' BlueOcean view, which truncates names....
Jenkinsfile:                // in Jenkins' BlueOcean view, which truncates names....
LICENSE:      form, that is based on (or derived from) the Work and for which the
LICENSE:      with the Work to which such Contribution(s) was submitted. If You
README.md:which can be found in the "rpm" and "deb" subdirectories, as well as scripts to
deb/build-deb:# - pkgRevision (usually "-0", see above), which allows updating packages with
deb/build-deb:#   we prefix the codename with a tilde (~), which effectively excludes it from
deb/build-deb:# EPOCH because of our use of CalVer, which made version comparing not work in
deb/common/rules:# zstd compression, which is non-standard, and breaks 'dpkg-sig --verify'
deb/gen-deb-ver:# non-pre-release versions, which would prevent users from updating from a pre-
deb/gen-deb-ver:        # be compared to determine which happened later, the commit hash identifes
deb/gen-deb-ver:                # Using BSD date (macOS), which doesn't support the --date option
rpm/SPECS/docker-ce.spec:# Provides modprobe, which we use to load br_netfilter if not loaded.
rpm/gen-rpm-ver:# - Version: > 1 : Only to be used for packaging-only changes (new package built for a version for which a package was already built/released)
rpm/gen-rpm-ver:# in this field, which is still allowed, but considered deprecated; see
rpm/gen-rpm-ver:# non-pre-release versions, which would prevent users from updating from a pre-
rpm/gen-rpm-ver:# > The tilde symbol (‘~’) is used before a version component which must sort
rpm/gen-rpm-ver:# > which wouldn’t otherwise sort appropriately.
rpm/gen-rpm-ver:        # be compared to determine which happened later, the commit hash identifes
rpm/gen-rpm-ver:                # Using BSD date (macOS), which doesn't support the --date option
static/gen-static-ver:  # be compared to determine which happened later, the commit hash identifes
static/gen-static-ver:          # Using BSD date (macOS), which doesn't support the --date option
verify:                         # may still be using "rawhide", which is not present on our package

git grep 'command '
deb/gen-deb-ver:        # as a pre-release before version v0.0.0, so that the go command prefers any
install-containerd-helpers:     if command -v dnf5; then
install-containerd-helpers:     elif command -v dnf; then
rpm/gen-rpm-ver:        # as a pre-release before version v0.0.0, so that the go command prefers any
static/gen-static-ver:  # as a pre-release before version v0.0.0, so that the go command prefers any
```

rpm/SPECS/docker-ce.spec
```diff
@@ -34,15 +34,13 @@ BuildRequires: cmake
 BuildRequires: gcc
 BuildRequires: git
 BuildRequires: glibc-static
-BuildRequires: libarchive
```
thaJeztah, line 37: This one was added in https://github.com/docker/docker-ce-packaging/commit/78242140d7f85863e8a8e3a133f3ebf73b9596ca, as part of https://github.com/docker/docker-ce-packaging/pull/554

It was to fix a failure when compiling tini;

> According to https://bugs.centos.org/view.php?id=18212, upgrading to libarchive-3.3.3-1.el8.x86_64 should resolve the problem.

Looks indeed that it's no longer needed (or at least, CI is happy), but perhaps I should make this a separate commit that's a clean _revert_ of the other one to help correlate / track down history.

thaJeztah (approved):
LGTM

let's bring this in

Issue #1012, opened 2024-04-10
/usr/bin/tar: Cannot change mode to rwxr-xr-x: Operation not permitted


AshwinHIBM:
**Description**
I'm getting the above error for a number of files with Fedora-40 only on ppc64le at https://github.com/docker/docker-ce-packaging/blob/master/rpm/Makefile#L44 when it runs the instruction `/usr/lib/rpm/rpmuncompress -x /root/rpmbuild/SOURCES/engine.tgz` only while running with the default seccomp profile.

**Steps to Reproduce**
Run the instruction
```bash
docker run --rm -e PLATFORM -v /workspace/docker-ce-packaging/rpm/rpmbuild/SOURCES:/root/rpmbuild/SOURCES:ro -v /workspace/docker-ce-packaging/rpm/rpmbuild/fedora-40/RPMS:/root/rpmbuild/RPMS -v /workspace/docker-ce-packaging/rpm/rpmbuild/fedora-40/SRPMS:/root/rpmbuild/SRPMS  rpmbuild-fedora-40/ppc64le -ba --define '_gitcommit_cli 2ae903e' --define '_gitcommit_engine 8b79278' --define '_release 1' --define '_version 26.0.0' --define '_origversion v26.0.0' --define '_buildx_rpm_version 0.13.1' --define '_buildx_version v0.13.1' --define '_buildx_gitcommit 7884339' --define '_compose_rpm_version 2.25.0' --define '_compose_version v2.25.0'  SPECS/docker-ce.spec SPECS/docker-ce-cli.spec SPECS/docker-ce-rootless-extras.spec SPECS/docker-buildx-plugin.spec SPECS/docker-compose-plugin.spec`
```

I tried replacing the default profile with a custom one including the system calls obtained from `strace` but that didn't help so there might be a missing call.

@thaJeztah @neersighted Can you please advise?

thaJeztah:
What version of docker are you running on? And with Fedora 40, you mean you're building _on_ Fedora 40, or building packages for Fedora 40 on your machine?

AshwinHIBM:
> What version of docker are you running on? And with Fedora 40, you mean you're building _on_ Fedora 40, or building packages for Fedora 40 on your machine?

Trying to build packages for Fedora-40 on our system for Docker v26.0.0. Sorry for the ambiguous wording.

thaJeztah:
[...]

PR #1007, opened 2024-03-19
[master] disable seccomp to prevent tar EPERM (Ubuntu 24.04 armhf on 20.04 host)

thaJeztah:
- addresses https://github.com/docker/docker-ce-packaging/pull/1006#issuecomment-2006878743

tar is failing to apply permissions, but only on `armhf` if seccomp is enabled

    docker run --rm -e PLATFORM -e EPOCH='5' -e DEB_VERSION=0.0.0~20240319100031.2356372 -e VERSION=0.0.0-20240319100031-2356372 -e CLI_GITCOMMIT=2356372 -e ENGINE_GITCOMMIT=23e1af4 -e BUILDX_VERSION=v0.13.1 -e BUILDX_DEB_VERSION=0.13.1 -e BUILDX_GITCOMMIT=7884339 -e COMPOSE_VERSION=v2.25.0 -e COMPOSE_DEB_VERSION=2.25.0 -v /home/ubuntu/workspace/docker-ce-packaging_PR-1006/deb/debbuild/ubuntu-noble:/build  debbuild-ubuntu-noble/aarch64
    + set -e
    + mkdir -p /root/build-deb/engine
    + tar -C /root/build-deb -xzf /sources/engine.tgz
    tar: engine/daemon/config: Cannot change mode to rwxrwxr-x: Operation not permitted
    tar: engine/daemon/stats: Cannot change mode to rwxrwxr-x: Operation not permitted
    tar: engine/daemon/images: Cannot change mode to rwxrwxr-x: Operation not permitted
    tar: engine/daemon/containerd: Cannot change mode to rwxrwxr-x: Operation not permitted
    ...
    tar: engine/integration-cli: Cannot change mode to rwxrwxr-x: Operation not permitted
    tar: Exiting with failure status due to previous errors
    make[1]: *** [Makefile:80: ubuntu-noble] Error 2
    make[1]: Leaving directory '/home/ubuntu/workspace/docker-ce-packaging_PR-1006/deb'
    make: *** [Makefile:83: ubuntu-noble] Error 2
    script returned exit code 2

This is failing when the ubuntu 24.04 armhf container runs on a ubuntu 20.04 host;

    Client: Docker Engine - Community
     Version:           25.0.3
     API version:       1.44
     Go version:        go1.21.6
     Git commit:        4debf41
     Built:             Tue Feb  6 21:22:23 2024
     OS/Arch:           linux/arm
     Context:           default

    Server: Docker Engine - Community
     Engine:
      Version:          25.0.3
[...]

PR #1013, opened 2024-04-12
[master] rpm: disable seccomp to prevent tar EPERM (Fedora 40 ppc64le on Debian)

thaJeztah:
- addresses / fixes https://github.com/docker/docker-ce-packaging/issues/1012
- relates to https://github.com/docker/docker-ce-packaging/pull/1007

commit 07b5d9031f18ad8317f23738c7ff4ce9f77e3ee3 disabled seccomp for the deb-builds because `chmod` failed tar was failing with an EPERM (Ubuntu 24.04 armhf on 20.04 host).

It looks like the same problem happens when running Fedora 40 on a Debian bookworm host on ppc64le.

That issue was confirmed to be related to `fchmodat2(2)`, which is a new syscall that's not supported on the host, and therefore gets the seccomp's default EPERM;

    fchmodat2(AT_FDCWD, "docker", 0775, AT_SYMLINK_NOFOLLOW) = -1 EPERM (Operation not permitted)
    tar: docker: Cannot change mode to rwxrwxr-x: Operation not permitted
    tar: Exiting with failure status due to previous errors

Ultimately we need to update the libseccomp profile to use ENOSYS (see moby ticket 42871) to allow it to degrade gracefully;

    fchmodat2(AT_FDCWD, "docker", 0775, AT_SYMLINK_NOFOLLOW) = -1 ENOSYS (Function not implemented)

### DIFF
--- a/rpm/Makefile
+++ b/rpm/Makefile
@@ -41,7 +41,13 @@ RPMBUILD_FLAGS?=-ba\
 
 # Additional flags may be necessary at some point
 RUN_FLAGS=
+
+# FIXME(thaJeztah): disabling seccomp to handle (ppc64le) tar "chown / chmod"
+# failing when running in a Fedora 40 container on a Debian bookworm host;
+# see https://github.com/docker/docker-ce-packaging/issues/1012 and
+# https://github.com/docker/docker-ce-packaging/pull/1006#issuecomment-2006878743
 RUN?=docker run --rm \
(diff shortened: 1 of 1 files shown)


thaJeztah:
cc @AshwinHIBM

vvoland (approved):

thaJeztah:
Let me bring this one in, but let us know if it worked, @AshwinHIBM

PR #1037, opened 2024-07-15
Remove 'rootlesskit-docker-proxy'

robmry:
**- What I did**

Depends on:
- https://github.com/moby/moby/pull/48132

With that PR, `rootlesskit-docker-proxy` is no longer used or built.

**- How I did it**

Removed commands for installing `rootlesskit-docker-proxy`.

**- How to verify it**

Interesting question!

**- Description for the changelog**

The executable `rootlesskit-docker-proxy` is no longer required and has been removed.

PR #578, opened 2021-09-09
[master] static: windows: fix "docker-proxy.exe" being copied as "dockerd.exe"

thaJeztah:
This was introduced in 09541b553c2f7646626efcc2797bd92ee48a66d0 (https://github.com/docker/docker-ce-packaging/pull/554), which fixed the source-location of the docker-proxy binary, but accidentally used "dockerd.exe" as target.

fixes https://github.com/moby/moby/issues/42609

PR #1045, opened 2024-08-06
[master] static: remove unused docker-proxy.exe binary from Windows packages

thaJeztah:
- relates to https://github.com/docker/docker-ce-packaging/pull/554
- relates to https://github.com/docker/docker-ce-packaging/pull/578
- slightly related to https://github.com/docker/docker-ce-packaging/pull/1037
- slightly related to https://github.com/moby/moby/pull/48132

The `docker-proxy.exe` binary was added to the static Windows packages through commit 09541b553c2f7646626efcc2797bd92ee48a66d0 (which introduced a bug, later fixed through fc5379fee869e8fa0e9a2569a66083e61e83de18).

It looks like that commit added the `docker-proxy.exe` binary with the assumption that it's used on Windows (given that its also built in upstream), but recent discussions revealed that's not the case.

This patch removes binary from the static packages for Windows, because it's not used.

A quick look at Docker Desktop's build scripts show that it's not included in those packages, and installation instructions for static binaries on Windows do not mention this binary https://docs.docker.com/engine/install/binaries/#install-server-and-client-binaries-on-windows), so no further changes should be needed.

```
Remove unused `docker-proxy.exe` binary from Windows packages
```

### DIFF
--- a/static/Makefile
+++ b/static/Makefile
@@ -85,7 +85,6 @@ cross-win: cross-win-engine
 	mkdir -p build/win/amd64/docker
 	cp $(CLI_DIR)/build/docker-windows-amd64.exe build/win/amd64/docker/docker.exe
 	cp $(ENGINE_DIR)/bundles/cross/win/dockerd.exe build/win/amd64/docker/dockerd.exe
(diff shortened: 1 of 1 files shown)


robmry (approved):
LGTM - thank you!

PR #1103, opened 2024-11-10
Add CentOS Stream 10.

Romain-Geissler-1A:
Use image "stream10-development" for now, after the release we should remove the "-development" suffix.


thaJeztah:
Thanks! This one will have to be carried to pick up the Jenkinsfile changes, and we need to have packages for containerd for CentOS 10 to be up on download.docker.com first.

PR #1108, opened 2024-11-14
[master] Use the final CentOS Stream 10 OCI image.

Romain-Geissler-1A:
@thaJeztah Yesterday RHEL 10 Beta was officially released, and it seems at the same time Red Hat has renamed the CentOS Stream 10 image name to it's final name. See https://quay.io/repository/centos/centos?tab=tags&tag=latest


thaJeztah (approved):
LGTM

tested locally

PR #1118, opened 2024-11-29
[master] deb,rpm: add kmod as recommended dependency to provide modprobe

thaJeztah:
- relates to https://github.com/moby/moby/pull/48992
- relates to https://github.com/moby/moby/pull/48492
- relates to https://github.com/moby/moby/pull/48960

Depending on the host configuration, the `br_netfilter` module may not be loaded by default. In this situation, docker will try to load the module through `modprobe`.

Older versions of docker would silently ignore failing to do so, log a message, and continue;

    time="2024-11-29T20:04:58.538404376Z" level=warning msg="Running modprobe bridge br_netfilter failed with message: , error: exec: \"modprobe\": executable file not found in $PATH"

However, starting with [db25b0dcd0461802289e962aa0df3abd323d1994][1] and [264c15bfc427d1321c5b302e48e16d113b06ef92][2], we now produce an error:

    === FAIL: libnetwork/drivers/bridge TestCreateFullOptions (0.04s)
    time="2024-11-29T19:03:44Z" level=error msg="Running modprobe br_netfilter failed with message: " error="exec: \"modprobe\": executable file not found in $PATH"
        bridge_linux_test.go:280: Failed to create bridge: loadBridgeNetFilterModule failed: cannot restrict inter-container communication: modprobe br_netfilter failed: exec: "modprobe": executable file not found in $PATH

This patch updates the packages to add `kmod` (which provides `modprobe`) as a recommended dependency. We should probably look for alternatives in our code to not require `modprobe`.

[1]: https://github.com/moby/moby/commit/db25b0dcd0461802289e962aa0df3abd323d1994
[2]: https://github.com/moby/moby/commit/264c15bfc427d1321c5b302e48e16d113b06ef92

**- What I did**

**- How I did it**

**- How to verify it**


austinvazquez (approved):

tianon (approved):

thaJeztah:
Thanks for reviewing! Since opening this PR, I also discussed a bit with @robmry @neersighted to look for possible alternatives that don't require `modprobe` (more so because running `modprobe` inside a container doesn't work for what we try to do.

I was a bit hesitant on adding a new dependency, but pending an alternative in the daemon itself, I think it's the most correct thing to do for the time being (i.e., we need `modprobe` in some cases, so we may as well reflect that in the package). 

Once we have an alternative for calling `modprobe`, this is probably fine to bring in (just need a reminder to "self" to remove  it again once we no longer need it.

Also opened a backport for 27.x;


- https://github.com/docker/docker-ce-packaging/pull/1121

robmry:
I just tried an equivalent to the `ip link show` trick, based on https://github.com/weaveworks/go-odp/blob/master/odp/dpif.go#L70-L90

It worked perfectly in a (privileged) docker-in-docker container on a Debian host that didn't have `ip6_tables` loaded. So I'll get a PR raised soon.

I'll make it try to fall-back to real `modprobe` ... so maybe it's still a good idea to add this dependency?

laurazard (approved):
LGTM

thaJeztah:
Thanks @robmry ! Yeah, I think this patch is still OK to have.

tianon:
Once we've got the `ip link show`-based codepath, we could probably downgrade from `Recommends:` to `Suggests:` instead 👀

thaJeztah:
Good callout! I actually think those changes are in for master / 28.0, so we already can!


Also: I think I  may have forgotten about "Suggests" being an option! (Maybe "suggests" would've been sufficient from the start?)

tianon:
IMO `Recommends:` is correct in the case where we explicitly shell out as our default case, but with a non-shell-out fallback in place, dropping to `Suggests:` is reasonable because it's much, _much_ less likely that we hit that codepath:

> #### `Recommends`
> This declares a strong, but not absolute, dependency.
>
> The `Recommends` field should list packages that would be found together with this one in all but unusual installations.

vs

> #### `Suggests`
> This is used to declare that one package may be more useful with one or more others. Using this field tells the packaging system and the user that the listed packages are related to this one and can perhaps enhance its usefulness, but that installing this one without them is perfectly reasonable.

(from https://www.debian.org/doc/debian-policy/ch-relationships.html#binary-dependencies-depends-recommends-suggests-enhances-pre-depends)

tianon:
Oh, and that was backported in 27.4.1+ (https://github.com/moby/moby/commit/8a8ab0d5670dd4f499422120ea4585d07ab464c0), so yeah, this is reasonable to downgrade to `Suggests` across the board.

PR #1124, opened 2024-12-11
deb, rpm: remove libseccomp-dev, libseccomp2, libselinux-devel, selinux-policy-devel, aufs-tools dependenciies

thaJeztah:
### deb, rpm: remove libseccomp-dev, libseccomp2 dependencies

We used to have `libseccomp-dev` as a build-time dependency, but this
part is handled by the OCI runtime (runc). Similarly, the  `libseccomp2`
runtime dependency is used by `runc`, not by `dockerd` directly.

The docker-ce package depends on `containerd.io`, which defines `libseccomp2`
as a dependency, so we can remove it from the docker-ce package;

    apt-cache depends containerd.io
    containerd.io
      Depends: libc6
      Depends: libseccomp2
      Conflicts: containerd
      Conflicts: runc
      Replaces: containerd
        containerd.io
      Replaces: runc
        containerd.io

- relates to https://github.com/moby/moby/pull/15832
- relates to https://github.com/moby/moby/pull/29264

### rpm: remove libselinux-devel dependency

The selinux code we use is purely a go implementation, and nothing links
libselinux.

### rpm: remove selinux-policy-devel

This dependency was added in [docker@42d8fe6], at which time we were building
a docker-selinux policy with additional policies, but we're now depending
on the distro's `container-selinux` package and it was removed in the moby
repository in [docker@bfe5cab], so we can remove it here as well. 

[docker@42d8fe6]: https://github.com/moby/moby/commit/42d8fe610548843376f035e1649444178f8497fb
[docker@bfe5cab]: https://github.com/moby/moby/commit/bfe5cab13bf0c83624fde0ab74f223bfa15d04c2

### deb: remove aufs-tools

The AuFS storage driver was deprecated and removed, so we can remove it
as a suggested dependency.

### rpm: remove version-constraint for container-selinux

The version constraint was added in e4b1c10a6b99b875cd5e37bbfef0b9a6ce58a518,
later updated in a246b19b0723d41a8e58ff42d3af52fd57a7e94d to prevent old
versions to be used, as they had some issues.

Current version on RHEL8 (which is the oldest rpm distro we build for) is
2:2.229, so we can remove this explicit constraint.

    Installing:
     container-selinux                noarch           2:2.229.0-2.module+el8.10.0+22417+2fb00970          ubi-8-appstream-rpms           71 k


thaJeztah:
cc @neersighted @tianon @vvoland ptal 🤗

neersighted (approved):

deb/common/control
```diff
@@ -39,7 +36,7 @@ Recommends: apparmor,
             pigz,
             procps,
             xz-utils
-Suggests: aufs-tools [amd64], cgroupfs-mount | cgroup-lite
```
thaJeztah, line 42: Double checked if it wasn't there for some other obscure purpose, but this was originally added as part of;

- https://github.com/moby/moby/pull/816

thaJeztah: ☝️ looks like there's potentially some other bits left behind in moby related to AuFS; I see [`auplink`](https://manpages.debian.org/jessie/aufs-tools/auplink.8.en.html) in there, and it looks like we still account for that in our AppArmor template; https://github.com/moby/moby/blob/7faa4ecfbfe5f641578931ff120c6e0f8faae16a/contrib/apparmor/template.go#L118-L136

thaJeztah:
Let me bring this one in; CI is happy 😄

PR #1138, opened 2025-01-07
add distro for opencloudos and tencentos  

BH1SCW:
**- What I did**

I add support for both OpenCloudOS and TencentOS , now we can build rpm packages for these two releases.
**- How I did it**

I made change in rpm/Makefile and add two folder for releases
**- How to verify it**

cd rpm
make  opencloudos
make tencentos 

we will get rpm packages in rpmbuild.

add distro for opencloudos and tencentos

### DIFF
--- a/rpm/Makefile
+++ b/rpm/Makefile
@@ -64,8 +64,10 @@ RUN?=docker run --rm \
 FEDORA_RELEASES ?= fedora-40 fedora-41
 CENTOS_RELEASES ?= centos-9 centos-10
 RHEL_RELEASES ?= rhel-8 rhel-9
+OPENCLOUDOS_RELEASES ?= opencloudos-9
+TENCENTOS_RELEASES ?= tencentos-4
 
-DISTROS := $(FEDORA_RELEASES) $(CENTOS_RELEASES) $(RHEL_RELEASES)
+DISTROS := $(FEDORA_RELEASES) $(CENTOS_RELEASES) $(RHEL_RELEASES) $(OPENCLOUDOS_RELEASES) $(TENCENTOS_RELEASES)
 BUNDLES := $(patsubst %,rpmbuild/bundles-ce-%-$(DPKG_ARCH).tar.gz,$(DISTROS))
 
 .PHONY: help
@@ -79,7 +81,7 @@ clean: ## remove build artifacts
 	-docker builder prune -f --filter until=24h
 
 .PHONY: rpm
-rpm: fedora centos ## build all rpm packages
+rpm: fedora centos opencloudos tencentos ## build all rpm packages
 
 .PHONY: fedora
 fedora: $(FEDORA_RELEASES) ## build all fedora rpm packages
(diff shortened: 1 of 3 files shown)


thaJeztah:
Thanks for contributing; I'm a bit on the fence on adding these currently, as there are no plans for docker to start building for these distros (and publish to download.docker.com).

That said, I'm looking at these Dockefiles, and IIUC, both these distros are a fork of / based on / really similar to the CentOS Dockerfiles, so I'm  wondering how far you would be able to get with the already present build-args;

I see that `DISTRO`. `SUITE`. and `BUILD_IMAGE` can be overridden currently; the remaining differences I see are;

Loading the GPG keys;

```dockerfile
RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-* || true
```

Instead of the hard-coded (CentOS 9 / 10);

```dockerfile
RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial
RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial-SHA256
```

Perhaps changing the CentOS ones to use a wildcard would work; not sure if the `|| true` is needed; that's worth checking

```dockerfile
RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-*
```

The second change appears to be the `powertools` repository (which was renamed in CentOS to `erb` I think?);

```dockerfile
RUN dnf config-manager --set-enabled crb || dnf config-manager --set-enabled powertools || true
```

versus (CentOS);

```dockerfile
RUN dnf config-manager --set-enabled crb
```

Perhaps it'd be ok to update the CentOS Dockerfile to do so if we add a comment describing that it's to allow re-purposing the Dockerfile for derrivatives, such as `opencloudos` or `tencentos`. For that one as well, it may be preferable to not have the `|| true` if we can do without, e.g.;

```dockerfile
# CentOS uses crb, but some derrivatives such as opencloudos and tencentos
# still use the old "powertools" name.
RUN dnf config-manager --set-enabled crb || dnf config-manager --set-enabled powertools
```

BH1SCW:
Thanks for your advice I have modiy my patch and I found this pr is already closed because I delete my branch,  so I create another pr base on this patch,

Could you help me to review it?

Best regards

BH1SCW:
I understand that you're a bit hesitant about accepting my patch. 
What I'd like to say is that as a developer, I'm more than willing to take responsibility for this project and maintain it in the long run. 
After all, this tool has tens of thousands of users. so, having the patch merged by upstream would be a more convenient way for maintenance, rather than forking a copy locally and maintaining two sets of code at the same time.

BH1SCW:
> Thanks for contributing; I'm a bit on the fence on adding these currently, as there are no plans for docker to start building for these distros (and publish to download.docker.com).
> 
> That said, I'm looking at these Dockefiles, and IIUC, both these distros are a fork of / based on / really similar to the CentOS Dockerfiles, so I'm wondering how far you would be able to get with the already present build-args;

I understand that you're a bit hesitant about accepting my patch.
What I'd like to say is that as a developer, I'm more than willing to take responsibility for this project and maintain it in the long run.
After all, this tool has tens of thousands of users. so, having the patch merged by upstream would be a more convenient way for maintenance, rather than forking a copy locally and maintaining two sets of code at the same time.

PR #1163, opened 2025-02-13
Bump compose v2.33.0

ndeloof:
**- What I did**

Bump compose v2.33.0
Declare buildx as a recommended dependency as v2.33.0 introduces support for delegating build command to bake.

**- How I did it**

**- How to verify it**

Compose v2.33.0

### DIFF
--- a/deb/common/control
+++ b/deb/common/control
@@ -104,6 +104,7 @@ Homepage: https://github.com/docker/buildx
 Package: docker-compose-plugin
 Priority: optional
 Architecture: linux-any
+Recommends: docker-buildx-plugin
 Enhances: docker-ce-cli
 Description: Docker Compose (V2) plugin for the Docker CLI.
  .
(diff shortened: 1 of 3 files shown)


vvoland (approved):
LGTM

deb/common/control
```diff
@@ -104,6 +104,7 @@ Homepage: https://github.com/docker/buildx
 Package: docker-compose-plugin
 Priority: optional
 Architecture: linux-any
+Recommends: docker-buildx-plugin
```
thaJeztah, line 107: Silly question; if compose shells out to buildx,  and running as “standalone” (`docker-compose`) does it shell out to `/whatever/path/for /plugins/docker-buildx` (standalone) or `docker buildx` ? Mostly looking at `Recommends: ` in the `docker-compose-plugin` package, and if that (when installed separate from the CLI) would also be able to use it, or if it should depend on `docker-ce-cli` already having the `Recommends:` for buildx

ndeloof: compose executes the CLI plugin directly (standalone) based on binary file path retrieved from `manager.GetPlugin`.

thaJeztah: Gotcha; thanks! Wanted to double-check we wouldn't end up in a case where the plug-in was there, but compose still trying to execute it in a different location.

We should probably look if we can have good ways for the CLI (when compose is executed as plugin) to pass information to the plugin(s), so that we have a single source of truth for resolving plugins, but that's an existing problem.

thaJeztah (approved):
LGTM

Issue #1186, opened 2025-04-15
manage docker group with systemd-sysusers


p5:
`systemd-sysusers` works from a declarative drop-in configuration file (e.g., /usr/lib/sysusers.d/myapp.conf) as an alternative to running groupadd inside the spec files.

This is especially useful in "atomic" distros, such as Silverblue, where groupadd is ran inside a CI environment and will not propagate to the end user system as /etc/passwd is owned by the system.

A better explanation is probably in the Fedora change document - https://fedoraproject.org/wiki/Changes/Adopting_sysusers.d_format

If the maintainers approve, I'd like to implement this for the RPMs.

tianon:
I think this is a great idea, and we should do it for the debs too. If the file is generic, it should maybe live in contrib in the Moby repo?

p5:
I am unfamilar with debs, but sysusers would be useful across the board.

It does require a systemd version newer than 215, which was released in July 2014 according to the Google Wikipedia summary.

The actual configuration would simply be a single line configuration file present in `/usr/lib/sysusers.d/*.conf` containing `g docker -`.  This says "I want a group (g) named docker with a dynamic GID".

p5:
I've made a couple PRs to demonstrate how this will look.

The [first](https://github.com/moby/moby/pull/49813) is ready to review in moby/moby.  This adds the configuration file which this packaging repo would be able to use.

The [second](https://github.com/docker/docker-ce-packaging/pull/1187) is an example of the changes required to implement here.  This one currently echos the file to the right location since I couldn't get the Makefile to use my fork's contents.

p5:
Hello
Any chance the [moby/moby](https://github.com/moby/moby/pull/49813) PR could be reviewed and/or merged?  I don't believe this will have _any_ affect on users since it will only be consumed when the packagers decide to.

I'd like to try and get sysusers in place soon for F43 (releasing in 5 months) to give it some time to brew.

PR #1195, opened 2025-05-20
feat: manage docker group with systemd-sysusers (carry #1187)

thaJeztah:
- carries / closes https://github.com/docker/docker-ce-packaging/pull/1187
- closes https://github.com/docker/docker-ce-packaging/issues/1186

Switches away from the groupadd postinstall commands to managing the docker group with sysusers.

This is a declarative way to create and manage users, better suited for the atomic distros such as Silverblue.

**- What I did**

**- How I did it**

**- How to verify it**

```markdown changelog
deb, rpm packages: manage docker group with systemd-sysusers instead of post-install script.
```


thaJeztah:
With this PR;

```bash
apt-get -y  update && apt-get -y install ca-certificates curl
install -m 0755 -d /etc/apt/keyrings
curl -fsSL https://download.docker.com/linux/debian/gpg -o /etc/apt/keyrings/docker.asc
chmod a+r /etc/apt/keyrings/docker.asc
echo   "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] https://download.docker.com/linux/debian $(. /etc/os-release && echo "$VERSION_CODENAME") stable" | tee /etc/apt/sources.list.d/docker.list > /dev/null
apt-get -y update
apt install ./docker-ce_0.0.0~20250520085713.af09051-1~debian.13~trixie_arm64.deb
```

The `docker.conf` is installed in `/etc/sysusers.d/docker.conf`

```bash
ls -l /etc/sysusers.d
total 4
-rw-r--r-- 1 root root 235 May 20 11:03 docker.conf
```


:question: https://manpages.debian.org/bookworm/debhelper/dh_installsysusers.1.en.html describes that `dh_installsysusers` would install this in `/usr/lib/sysusers.d/docker.conf` - is that the more correct location for this? @tianon @neersighted ?

tianon:
If it's installed by a package, yes, absolutely.

thaJeztah:
> If it's installed by a package, yes, absolutely.

Updated to install in `/usr/lib/sysusers.d/`


I was also still trying to use the symlink approach, but didn't manage to get that working so far; https://github.com/docker/docker-ce-packaging/pull/1196

thaJeztah:
@vvoland @tianon @neersighted ptal

p5 (approved):
Thank you for taking over this PR!
Work has got extremely busy lately and I've not had chance to do much else.

These changes all look great, and all make complete sense.

thaJeztah:
> Work has got extremely busy lately and I've not had chance to do much else.

No worries!!


> I too couldn't figure out the Debian dh_installsysusers part, so did what I thought was the bare minimum to get something working, bypassing that.


Yeah, same; decided to keep it as a follow-up for now (but likely also requires https://github.com/docker/docker-ce-packaging/pull/1202, which also seems to have issues still to look into).

PR #1210, opened 2025-05-28
fix(all): revert sysusers

p5:
**- What I did**

Reverted the systemd-sysusers change.

```markdown changelog

```

### DIFF
--- a/deb/common/docker-ce.postinst
+++ b/deb/common/docker-ce.postinst
@@ -1,0 +1,20 @@
+#!/bin/sh
+set -e
+
+case "$1" in
+	configure)
+		if [ -z "$2" ]; then
+			if ! getent group docker > /dev/null; then
+				groupadd --system docker
+			fi
+		fi
+		;;
+	abort-*)
+		# How'd we get here??
+		exit 1
+		;;
+	*)
+		;;
+esac
+
+#DEBHELPER#
--- a/deb/common/rules
+++ b/deb/common/rules
@@ -133,9 +133,6 @@ override_dh_auto_install:
 	install -D -p -m 0755 engine/contrib/dockerd-rootless-setuptool.sh debian/docker-ce-rootless-extras/usr/bin/dockerd-rootless-setuptool.sh
 	# TODO: how can we install vpnkit?
 
-	# install systemd sysusers config
-	install -D -p -m 0644 engine/contrib/systemd-sysusers/docker.conf debian/docker-ce/usr/lib/sysusers.d/docker.conf
-
 override_dh_installinit:
 	# use "docker" as our service name, not "docker-ce"
 	dh_installinit --name=docker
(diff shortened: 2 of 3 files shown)


thaJeztah:
Thanks! /cc @vvoland FYI


We'll probably be rebuilding the packages without the systemd-sysusers patch, and deploy new packages, keeping the systemd-sysusers change for a next release

p5:
@thaJeztah 
I'm honestly thinking revert everything.  I clearly don't understand this enough, and I wasn't sure my "fix" even fixed it.

thaJeztah:
Yes; that's what we want to do. looks like it's the last commit in the branch, so we can build using the commit before that (but we could do an explicit revert PR as well)

p5:
Ok - please do whatever you can to fix the immediate issue of the broken release, and I'll figure out the proper fix to the issue.  

I'm kinda strongly for reverting at least the Debian scripts since I'm completely in the dark on how it works, and it seems there's quite a bit of tech debt in this area.

thaJeztah:
No worries! We're also looking why our deploy pipeline didn't flag it; likely because it may have verified an upgrade on the same machine, so the user/group to have been there.

renanferreiratrimble (approved):

p5:
I'm thinking the "proper" fix (i.e. to keep using sysusers across the board) would be to do something like:

```
diff --git a/deb/common/docker-ce.postinst b/deb/common/docker-ce.postinst
new file mode 100755
index 0000000..5bf9251
--- /dev/null
+++ b/deb/common/docker-ce.postinst
@@ -0,0 +1,20 @@
+#!/bin/sh
+set -e
+
+case "$1" in
+       configure)
+               if [ -z "$2" ]; then
+                       if ! getent group docker > /dev/null; then
+                               systemd-sysusers docker.conf
+                       fi
+               fi
+               ;;
+       abort-*)
+               # How'd we get here??
+               exit 1
+               ;;
+       *)
+               ;;
+esac
+
+#DEBHELPER#
diff --git a/rpm/SPECS/docker-ce.spec b/rpm/SPECS/docker-ce.spec
index 57d8a90..2a785d9 100644
--- a/rpm/SPECS/docker-ce.spec
+++ b/rpm/SPECS/docker-ce.spec
@@ -107,6 +107,9 @@ mkdir -p ${RPM_BUILD_ROOT}/etc/docker
 %{_mandir}/man*/*
 %dir /etc/docker
 
+%pre
+%systemd_sysusers_pre docker.conf
+
 %post
 %systemd_post docker.service
```

This adds back the postinstall command to the deb, which invokes the sysusers file.  And uses a built-in macro for RPMs.

Though in the other thread, somebody said sysusers isn't used in Ubuntu and Debian, which I wasn't aware of.  How do you want to play this?

vvoland (approved):

PR #1212, opened 2025-06-05
Add `-w` to buildx ldflags

tianon:
This disables DWARF generation, which dramatically decreases the size of the binary.

Notably, this does *not* include `-s` which disables the symbol table, as doing so would *also* make it so that `govulncheck` cannot be run on the resulting binary with meaningful results.

Before is ~90M and after is ~72M.

If we add `-s` as well, that ~72M drops to ~63M, but we also lose `govulncheck` so again, IMO, that's not worth doing.

See also:
- https://github.com/docker/compose/pull/10325
- https://github.com/docker/buildx/blob/c45185fde0c9e596afb9044be3328b6d39b3d7ca/Dockerfile#L86

### DIFF
--- a/deb/common/rules
+++ b/deb/common/rules
@@ -40,7 +40,7 @@ override_dh_auto_build:
 			go build \
 				-mod=vendor \
 				-trimpath \
-				-ldflags "-X github.com/docker/buildx/version.Version=$(BUILDX_VERSION) -X github.com/docker/buildx/version.Revision=$(BUILDX_GITCOMMIT) -X github.com/docker/buildx/version.Package=github.com/docker/buildx" \
+				-ldflags "-w -X github.com/docker/buildx/version.Version=$(BUILDX_VERSION) -X github.com/docker/buildx/version.Revision=$(BUILDX_GITCOMMIT) -X github.com/docker/buildx/version.Package=github.com/docker/buildx" \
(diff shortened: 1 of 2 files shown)


thaJeztah:
Is this something we can / should do for other binaries as well?

oh; recalls me that I tried to update to compat 13, which had a todo that an override could be removed, but that doesn't seem to work 😅

- https://github.com/docker/docker-ce-packaging/pull/1202

thaJeztah (approved):
SGTM


but asked Tonis and Kevin to have a look

tianon:
It is something you could/should do for other binaries, but I think they're all using upstream build scripts, so those changes don't happen in this repository.  (For Compose in particular, this already happened back in https://github.com/docker/compose/pull/10325, for example.)

tonistiigi:
> but asked Tonis and Kevin to have a look

We already do this in upstream iiuc

tianon:
> We already do this in upstream iiuc

Yep, quoted in my OP: :sweat_smile:

> * https://github.com/docker/buildx/blob/c45185fde0c9e596afb9044be3328b6d39b3d7ca/Dockerfile#L86

(IMO that should drop `-s` so that `govulncheck` can work on the binaries, but that's a separate discussion)

thaJeztah:
> We already do this in upstream iiuc

I'll assume that's a "LGTM", so let me merge this one.

PR #1218, opened 2025-06-18
bump Model version to v0.1.30

xenoscopic:
**- What I did**

I bumped the Docker Model CLI plugin to v0.1.30.

This fixes a race condition with DMR installs on Docker CE and Cloud when multiple Compose model providers are specified.  It also fixes status displays with Compose model providers.


thaJeztah (approved):
LGTM

I added a short release note

xenoscopic:
Maybe we should wait on https://github.com/docker/model-cli/pull/98 and do another tag; I'd prefer to fix these things the right way.  But we do need to have the fix in this CE release either way.  @thaJeztah or @vvoland do you know when the final cutoff date is?

PR #1223, opened 2025-07-04
rpm/deb: remove libtool-ltdl build dependency

vvoland:
The Docker CLI no longer requires libtool-ltdl as a build dependency.

Previously it was an indirect dependency introduced by pkcs11, but it's
no longer needed after:
https://github.com/docker/cli/commit/cb3e55bf5802550428867ee8eba0f1a8e3212813

This patch removes these dependencies from both the Debian control file
and RPM spec files.

Signed-off-by: Paweł Gronowski <pawel.gronowski@docker.com>


crazy-max (approved):

vvoland:
Jenkins failures are just flakes (`java.io.StreamCorruptedException: invalid stream header: 636F7272`)

crazy-max (approved):

PR #1249, opened 2025-09-12
Add Fedora 43.

Romain-Geissler-1A:
Blindly copy the fedora 42 Dockerfile for fedora 43, just change the version number.

```markdown changelog
Add Fedora 43 packages
```

### DIFF
--- a/distros.json
+++ b/distros.json
@@ -58,7 +58,15 @@
             "amd64",
             "aarch64"
         ],
-        "end_of_life": "November, 2025"
+        "end_of_life": "May 13, 2026"
+    },
+    "fedora-43": {
+        "image": "fedora:43",
+        "arches": [
+            "amd64",
+            "aarch64"
+        ],
+        "end_of_life": "December 02, 2026"
     },
     "raspbian-bullseye": {
         "image": "balenalib/rpi-raspbian:bullseye",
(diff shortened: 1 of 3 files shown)


Romain-Geissler-1A:
In the past we have merged Fedora N support before its actual release (usually around the beta phase where things are stabilizing). If you mark this "not stable" then on the D day of the Fedora release packages won't be available (even after merging this, no new docker/containerd package will be made available until the next containerd/docker actual release, which on occasion can take some time).

I try to proactively send pull requests like these since couple of years since I regularly hit the issue that docker (I am especially interested in the docker client) was not installable from the `quay.io/fedora/fedora` OCI image more or less every 6 months and that created issues in my own CIs.

vvoland (dismissed):

vvoland:
@Romain-Geissler-1A  do you have a moment to rebase this one? `Jenkinsfile` is gone now.

Otherwise I can rebase this one and get it in.

vvoland (approved):

Romain-Geissler-1A:
Thanks for the rebase, I was off yesterday.

So, don't we need https://github.com/docker/containerd-packaging/pull/427 as well ? Because the resulting docker rpm packages won't be installable without the required containerd dependency.

PR #1255, opened 2025-09-24
rpm: Add iptables-nft requirement for Fedora 43+ and RHEL 10+

vvoland:
Fedora 43 and RHEL 10 have transitioned to nftables as the default packet filtering framework.

By default Docker requires iptables compatibility to function properly with these systems.


thaJeztah (approved):
SGTM


cc @robmry to double-check ❤️

vvoland:
Keeping as a draft so we don't merge before 28.5.0

PR #1259, opened 2025-10-03
bump compose to v2.40.0

ndeloof:
**- What I did**
bump compose to v2.40.0

```markdown changelog
Docker Compose v2.40.0
```

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -41,7 +41,7 @@ DOCKER_CLI_REF     ?= $(REF)
 DOCKER_ENGINE_REF  ?= $(REF)
 # DOCKER_COMPOSE_REF is the version of compose to package. It usually is a tag,
 # but can be a valid git reference in DOCKER_COMPOSE_REPO.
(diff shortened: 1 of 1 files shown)


common.mk
```diff
@@ -41,7 +41,7 @@ DOCKER_CLI_REF     ?= $(REF)
 DOCKER_ENGINE_REF  ?= $(REF)
 # DOCKER_COMPOSE_REF is the version of compose to package. It usually is a tag,
 # but can be a valid git reference in DOCKER_COMPOSE_REPO.
-DOCKER_COMPOSE_REF ?= v2.39.4
+DOCKER_COMPOSE_REF ?= v2.40.9
```
thaJeztah, line 44: `.0` probably 😅 


```
Fatal: couldn't find remote ref v2.40.9
```

thaJeztah: Yeah, discussion already happening on that; @vvoland where does this need to be updated in `docker/packaging` ?

PR #1263, opened 2025-10-06
RPM: require "iptables-nft or iptables"

robmry:
**- What I did**

Rather than requiring iptables-nft for RHEL10/Fedora43 and iptables for other versions and CentOS, use "(iptables-nft or iptables)".

Newer OSs will have iptables-nft, if anything - and making people install iptables-legacy isn't great if it's not needed.

(When neither dependency is available - dnf seems to install both if they're available, but doesn't mind if it can only find one of them. And, if one's already installed, it doesn't try to install the other.)

```markdown changelog

```

### DIFF
--- a/rpm/SPECS/docker-ce.spec
+++ b/rpm/SPECS/docker-ce.spec
@@ -19,12 +19,7 @@ Requires: docker-ce-cli
 Recommends: docker-ce-rootless-extras
 Requires: container-selinux
 Requires: systemd
-%if (0%{?fedora} >= 43) || (0%{?rhel} >= 10)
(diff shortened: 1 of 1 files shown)


vvoland (approved):
LGTM

For a follow up, I think we could do the same (`iptables | nftables`) for debs, right?

https://github.com/docker/docker-ce-packaging/blob/4fe9f66b0e17f98e7ce864b72aacec018b2e512a/deb/common/control#L29-L30

robmry:
> For a follow up, I think we could do the same (`iptables | nftables`) for debs, right?

I looked in to that, it doesn't seem to be needed because on Debian the `iptables` dependency ends up installing `iptables-nft`.

PR #1309, opened 2026-04-07
rootless: drop dependency on fuse-overlayfs

AkihiroSuda:
**- What I did**

rootless: drop dependency on fuse-overlayfs.

fuse-overlayfs is no longer needed since kernel 5.11. EL 8 (kernel 4.18) distros still need it, but official RPMs are no longer built for them.
https://download.docker.com/linux/centos/8/x86_64/stable/Packages/

```markdown changelog
rootless: drop dependency on fuse-overlayfs
```


vvoland (approved):
LGTM, thanks!

For a follow up, we should also adjust: https://github.com/docker/packaging/blob/e3a290fab82446207f45a4142887be348c0cddc5/pkg/docker-engine/rpm/docker-ce-rootless-extras.spec#L20

PR #1318, opened 2026-05-05
rpm: Ship SELinux policy denying AF_ALG sockets

vvoland:
**Note: This requires dockerd to be configured with`selinux-enabled: true` (through `daemon.json` or cli flag). It is NOT enabled by default**

Ship a CIL policy module that denies AF_ALG socket creation for all container domains (container_t, spc_t). Like AppArmor, SELinux hooks into the kernel's security_socket_create() LSM callback, which fires regardless of whether the socket was created via socket(2) or the legacy socketcall(2) multiplexer. This blocks AF_ALG selectively without disrupting other socketcall usage by 32-bit binaries.

This complements the seccomp profile change in moby, which blocks socket(AF_ALG) but cannot filter socketcall arguments. On SELinux-only systems (Fedora, RHEL, CentOS) where AppArmor is not available, this CIL module provides equivalent protection.

Load the module from %post when SELinux is enabled, warning but keeping installation non-fatal if semodule cannot load it. Remove the module on uninstall.

**- What I did**

```markdown changelog

```

### DIFF
--- a/rpm/SPECS/docker-ce.spec
+++ b/rpm/SPECS/docker-ce.spec
@@ -104,6 +104,9 @@ install -D -p -m 0644 engine/contrib/init/systemd/docker.socket ${RPM_BUILD_ROOT
 # install manpages
 make -C ${RPM_BUILD_DIR}/src/engine/man DESTDIR=${RPM_BUILD_ROOT} mandir=%{_mandir} install
 
+# install SELinux policy to deny AF_ALG sockets in container domains
+install -D -m 644 engine/contrib/selinux/docker-af-alg-deny.cil %{buildroot}%{_datadir}/docker-ce/selinux/docker-af-alg-deny.cil
+
 # create the config directory
 mkdir -p ${RPM_BUILD_ROOT}/etc/docker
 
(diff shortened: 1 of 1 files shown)


rpm/selinux/docker-af-alg-deny.cil
```diff
@@ -0,0 +1,12 @@
...
+;; legitimate use in containers and enables trivial privilege escalation
+;; on unpatched kernels.
+;;
+;; This rule targets container_domain (container_t, spc_t, etc.) and
+;; does not affect the container runtime (container_runtime_t) or host
+;; processes.
+;;
+;; Requires SELinux userspace >= 3.6 for CIL deny support.
+;; Requires container-selinux for the container_domain attribute.
+(deny container_domain self (alg_socket (create)))
```
thaJeztah: would we want to add this in https://github.com/moby/moby/tree/master/contrib/selinux ?

vvoland: Ah yes, that would work!

thaJeztah (approved):
LGTM

PR #1158, opened 2025-01-24
[master] deb, rpm: make kmod "suggests" instead of "recommends"

thaJeztah:
- updates, fixes https://github.com/docker/docker-ce-packaging/pull/1118#issuecomment-2611224201

This was added in deed8d9df8cd90139229748021efcc006900aba5, as the docker engine depended on modprobe to enable br_netfilter. Docker Engine no longer requires this since [moby/moby@4b8c720], and [moby/moby@8a8ab0d] (docker\ engine v27.4.1), so we can reduce this to a "suggests"

[moby/moby@4b8c720]: https://github.com/moby/moby/commit/4b8c72060d1089626e4ac9f5e0624e0cb567d4bc
[moby/moby@8a8ab0d]: https://github.com/moby/moby/commit/8a8ab0d5670dd4f499422120ea4585d07ab464c0

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #1331, opened 2026-06-11
deb, rpm: remove kmod dependency

thaJeztah:
relates to:

- https://github.com/moby/moby/pull/49038
- https://github.com/docker/docker-ce-packaging/pull/1118
- https://github.com/docker/docker-ce-packaging/pull/1158

This was added in deed8d9df8cd90139229748021efcc006900aba5, as the docker engine depended on modprobe to enable br_netfilter. Docker Engine no longer requires this since [moby@4b8c720], so we can remove this.

[moby@4b8c720]: https://github.com/moby/moby/commit/4b8c72060d1089626e4ac9f5e0624e0cb567d4bc

**- What I did**

```markdown changelog

```

### DIFF
--- a/deb/common/control
+++ b/deb/common/control
@@ -36,8 +36,7 @@ Recommends: apparmor,
             pigz,
             procps,
             xz-utils
-Suggests: cgroupfs-mount | cgroup-lite,
-            kmod,
+Suggests: cgroupfs-mount | cgroup-lite
 Conflicts: docker.io
 Replaces: docker-ce-cli (<< 5:28.0.0)
 Description: Docker: the open-source application container engine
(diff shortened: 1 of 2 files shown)


vvoland (commented):
Does the non-modprobe way work reliably? Or is there still a chance it will need to fallback to modprobe?

robmry:
> Does the non-modprobe way work reliably? Or is there still a chance it will need to fallback to modprobe?

I guess we can't tell whether the fallback has been used, but I think non-modprobe is reliable.

If modprobe is really needed, "suggests" probably isn't going to help much anyway - so I think the change is good.

robmry (approved):

PR #1350, opened 2026-09-03
rpm: Apply AF_VSOCK SELinux policy

vvoland:
- needs: https://github.com/moby/moby/pull/53551

Activate the known socket-deny CIL modules from one shared list so install and final uninstall remain symmetric. Keep unsupported modules non-fatal on older SELinux userspace.

Use the filename-derived module names during removal, which also fixes the stale AF_ALG removal name.

**- What I did**

```markdown changelog

```


thaJeztah (approved):
LGTM

